PR #3, opened 2019-01-15
REVAI-136: Clean up SDK

kbridbur:
Adds class based responses and error parsing
Cleans up method parameters to match spec

Everything should be well linted for the most part. Some comments go a few characters beyond 80 but I thought it would be fine to let those slide


src/rev_ai/models/error.py
```diff
@@ -0,0 +1,112 @@
+class ProblemDetail:
+	def __init__(
```
amikofalvy: Should we use tabs or spaces? Ask @JDongian.

src/rev_ai/speechrec.py
```diff
@@ -9,67 +9,85 @@
 import requests
 import logging
 import json
+from models import *
 
 LOG = logging.getLogger(__name__)
 
+class RevSpeechAPIClient:
```
kennygea: is this still RevSpeech? or should this become `RevAiApiClient`

kennygea: I could for see that this client has more than 1 API under it

src/rev_ai/speechrec.py
```diff
@@ -9,67 +9,85 @@
...
         self.s.headers.update({
             'Authorization': 'Bearer {api_key}'.format(api_key=api_key)
         })
-        if use_stage:
-            self.BASE_URL = ("https://api-stage.rev.ai/revspeech/{v}/"
-                             .format(v=self.VERSION))
 
 
-    def submit_job_url(self, media_url, metadata="", callback_url=None):
+    def submit_job_url(
```
kennygea: question, should we have 2 urls for url/local file? 

similar for get _transcript, I would think we should have 1 method and allow the user to specify the type output which is a enum/string which is code somewhere.

I don't feel very strongly about submit_job, but I think get_transcript should be one method (which gives these methods a 1-1 relationship with available endpoints

src/rev_ai/speechrec.py
```diff
@@ -9,67 +9,85 @@
...
+            payload['callback_url'] = options.callback_url
         response = self.s.post(url_jobs, json=payload)
 
-        return response.json()
-
-    def submit_job_local_file(self, filename,
-                              media_type="audio", content_type=None, metadata="", callback_url=None):
+        if (response.status_code == 200):
+            return Job.from_json(response.json())
+        elif (response.status_code == 400):
```
kennygea: what about 401 errors. Should we catch them on creation of the client? or on the method?

kennygea: I guess then what about 5xx responses? I'd imagine `ApiError.from_json` isn't going to work

src/rev_ai/models/error.py
JDongian: spaces.
The same arguments apply for any language (except Make, where it is not an option)
* pastability
* everyone does it
* easier to align funky function arguments

JDongian:
https://github.com/revdotcom/revai-python-sdk/blob/master/Makefile#L53
`make lint`
you will need to install flake8: `pip install flake8`

Also, why is it called an SDK? Feels very foreign. Python has packages and modules. SDKs are generally big libraries with C code linked in. Even numpy is a "python package". The first python SDK off the top of my mind is the "google cloud python SDK", because it does _a lot_.

src/rev_ai/models/__init__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+"""Models"""
+
+from models.job import Job
+from models.job_options import JobSubmitOptions
+from models.error import ApiError, InsufficientBalanceError, InvalidParametersError, InvalidValueError
```
JDongian: if you want to stay in 80 char limits (very doable in python), you can use the tuple with explicit parenthesis pattern (as you use below)

src/rev_ai/models/__init__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
```
JDongian: I like this, try to get it in all files. It helps on some weird Windows systems with funky encodings. I've had people not be able to install my modules because I forgot this.

src/rev_ai/models/error.py
```diff
@@ -0,0 +1,112 @@
...
+		type_,
+		title,
+		detail
+	):
+		self.status = status
+		self.type = type_
+		self.title = title
+		self.detail = detail
+
+class ApiError:
```
JDongian: https://docs.python.org/2/tutorial/errors.html#tut-userexceptions

JDongian: the only related difference in python3 is that there is some (here irrelevant) syntactic sugar for certain class relationships

src/rev_ai/models/job_options.py
```diff
@@ -0,0 +1,14 @@
...
+		self, 
+		metadata=None, 
+		callback_url=None, 
+		skip_diarization=None
+	):
+		"""
+		:param metadata: info to associate with the transcription job
+        :param callback_url: callback url to invoke on job completion as a webhook
+        """
+		self.metadata = "" if metadata == None else metadata
```
JDongian: `is None` (flake8 will find this for you)

src/rev_ai/models/job_options.py
```diff
@@ -0,0 +1,14 @@
...
+		self.callback_url = callback_url
+		self.skip_diarization = False if skip_diarization == None else skip_diarization
```
JDongian: this is a bit weird for me, shouldn't this be a `bool`? In which case the right hand expression should simply be `skip_diarization is not None`.

src/rev_ai/speechrec.py
```diff
@@ -9,67 +9,85 @@
...
+            return InvalidParametersError.from_json(response.json())
```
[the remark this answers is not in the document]
JDongian: An example, so you don't have to take my word for it:
https://github.com/boto/boto3/blob/develop/boto3/s3/transfer.py#L285

src/rev_ai/models/account.py
```diff
@@ -0,0 +1,18 @@
+class Account:
+	def __init__(
+		self,
```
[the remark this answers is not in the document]
JDongian: especially for lightweight classes, this makes more sense I think.

src/rev_ai/models/job.py
```diff
@@ -0,0 +1,48 @@
+class Job:
+	def __init__(
+		self,
+		id_,
```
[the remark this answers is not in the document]
JDongian: Name collision. `id` is a python keyword. Helps the syntax highlighter a bit.

Note that people often disregard some python keywords, such as `file`, `id`, and `type`, overwriting them at will (happens in some well known packages). So it's just a matter of taste.

src/rev_ai/speechrec.py
```diff
@@ -9,67 +9,85 @@
...
 
+class RevSpeechAPIClient:
+    """Client which implements rev.ai API"""
+
+    VERSION = "v1"
+    """Default version of rev.ai"""
 
-class RevSpeechAPI:
-    VERSION = "v1beta"
     BASE_URL = "https://api.rev.ai/revspeech/{v}/".format(v=VERSION)
```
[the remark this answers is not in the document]
JDongian: api.rev.ai/translate/v1 seems cool

src/rev_ai/models/error.py
```diff
@@ -0,0 +1,112 @@
...
+	):
+		self.problem_details = problem_details
+
+	@classmethod
+	def from_json(
+		cls,
+		json
+	):
+		details = ProblemDetail(
+			json.get("status"),
```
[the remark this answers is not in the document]
JDongian:
```
In [1]: d = {1: 2, 2: 4, 100: 200, 111: 222}                                                       

In [2]: d.get(1)                                                                                   
Out[2]: 2

In [3]: d.get(3)                                                                                   

In [4]: d[1]                                                                                       
Out[4]: 2

In [5]: d[3]                                                                                       
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-5-0acadf17a380> in <module>
----> 1 d[3]

KeyError: 3

```
`.get` has a implied default value of `None`.

src/rev_ai/speechrec.py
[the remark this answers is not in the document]
pjhuck: yeah, i'd be in favor of changing the url. `speech/job` seems a little broad. Possible suggestions `transcribe` or `speechtotext`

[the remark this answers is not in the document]
dkokotov: I like `speechtotext`

src/rev_ai/speechrec.py
JDongian: I would say two methods is preferable to the above. I would avoid function signatures where we want the user to set `media_url` XOR `file`. Another option is to 'smartly detect' if the provided media is a file on the system. But my most preferred is with two separate functions.

src/rev_ai/speechrec.py
```diff
@@ -91,33 +119,63 @@ def view_job(self, id_):
...
+
+        if (response.status_code == 200):
+            return response.text
+        elif (response.status_code == 406 or response.status_code == 409):
+            return InvalidValueError.from_json(response.json())
         else:
-            content_type_accept = 'text/plain'
+            return ApiError.from_json(response.json())
+
+    def get_transcript_as_json(
```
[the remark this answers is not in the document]
JDongian: I think I prefer the latter, in which case we would change the name to `_as_transcript` or something similar

src/rev_ai/speechrec.py
kennygea: Yeah for submit job, it makes sense to have 2.

I would have 1 for get transcript though

src/rev_ai/models/account.py
```diff
@@ -0,0 +1,18 @@
...
+		email,
+		balance_seconds
+	):
+		self.email = email
+		self.balance_seconds = balance_seconds
+
+	@classmethod
+	def from_json(
```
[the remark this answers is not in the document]
kbridbur: Standard seems to be to use a mix, and indent like this when the method def would be too long 
Ex. https://github.com/googleapis/google-cloud-python/blob/master/speech/google/cloud/speech_v1/helpers.py

src/rev_ai/models/error.py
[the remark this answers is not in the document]
kbridbur: As Josh says .get has a default value where [ ] raises an error when the key doesnt exist. This was my way of doing assertions that certain properties existed. 

Though i can see an argument for just doing .get for everything for consistency

src/rev_ai/speechrec.py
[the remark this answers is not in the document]
kbridbur: :+1:

src/rev_ai/models/error.py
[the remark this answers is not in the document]
dkokotov: i think its fine to have both, it makes sense

src/rev_ai/models/job.py
```diff
@@ -0,0 +1,48 @@
+class Job:
+	def __init__(
```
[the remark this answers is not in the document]
kbridbur: I skimped a bit on docs in the beginning while a remembered how python is written. I'll be adding docs in a pass later

src/rev_ai/models/account.py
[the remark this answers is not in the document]
dkokotov: yeah makes sense. but this method list is short enough i think to put it on one line.

src/rev_ai/models/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+"""Models"""
+
+from models.job import Job
+from models.job_options import JobSubmitOptions
+from models.error import ApiError, InsufficientBalanceError, InvalidParametersError, InvalidValueError
+from models.account import Account
+
+__all__ = (
```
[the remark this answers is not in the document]
kbridbur: Though, based on a few of Josh's comments I think I will be removing it here

src/rev_ai/speechrec.py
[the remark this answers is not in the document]
dkokotov: filed https://bugs.rev.com/browse/REVAI-143

src/rev_ai/__init__.py
```diff
@@ -1,4 +1,4 @@
 # -*- coding: utf-8 -*-
 """Top-level package for rev_ai"""
 
-__version__ = '0.0.4'
```
JDongian: https://semver.org/
this isn't back compatible

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,130 @@
...
+import logging
+import json
+from models import (
+    Job,
+    JobSubmitOptions,
+    Account,
+    Transcript
+)
+
+LOG = logging.getLogger(__name__)
```
JDongian: unused, delete?
usage:
`LOG.info("sending job")`
`LOG.critical("server not found")`

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,130 @@
...
+
+class RevAiAPIClient:
+    """Client which implements rev.ai API"""
+
+    version = "v1"
+    """Default version of rev.ai"""
```
JDongian: * i've never seen a comment below the relevant line
* usually don't use triple quote unless quoting a multiline string or method docstring. I would have used `#`

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,130 @@
...
+"""
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
+import os
+import requests
+import logging
+import json
+from models import (
```
JDongian: > I was hoping to find a better way to import the classes I wanted using the init.py in the models folder. Is there none?

https://www.python.org/dev/peps/pep-0020/ says:
Explicit is better than implicit.

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,130 @@
...
+
+        response = self.s.get(url_jobs_id)
+
+        return Job.from_json(response.json())
+
+    def get_transcript(self, id_, format_):
+        """Get the transcript of a specific job as json.
+
+        :param id_: id that the server gave you
+        :param format_: format of returned transcript, either "text" or "json"
```
JDongian: I'm not the biggest fan of magic string inputs. Documentation helps, but better is to have the function signature make things obvious. My personal choice would be a `fmt_text=True` flag (which wouldn't work if you have more than two formats). If more than two formats are planned, I would make two methods now.

Python enums suck.

kbridbur: 😢Makes sense, I was hoping to have an enum and it sucked. If we're going to have a boolean I would rather just have 2 methods

src/rev_ai/speechrec.py
kbridbur: As part of a conversation with Josh it became clear that the standard in python is to write the method without errors in mind and let them pop up as actual errors instead of seemingly valid responses. We can talk about this more if you'd like.

src/rev_ai/speechrec.py
kbridbur: I think since get transcript has multiple return types it makes sense to have 2 methods. Yes this breaks the 1:1 endpoint:method relationship we had with the API but I think the SDK if more clear because of it

src/rev_ai/models/error.py
kbridbur: Everything should be spaces now. :+1:

src/rev_ai/__init__.py
```diff
@@ -1,4 +1,4 @@
...
+__version__ = '0.1.0'
```
kbridbur: As Josh noted, these are breaking changes

amikofalvy: Josh is correct. But I think this requires a bump in the major version.
```suggestion
__version__ = '1.0.0'
```

Relevant Docs: https://semver.org/

> - MAJOR version when you make incompatible API changes,
> - MINOR version when you add functionality in a backwards-compatible manner, and
> - PATCH version when you make backwards-compatible bug fixes.

setup.py
```diff
@@ -37,8 +37,8 @@
     version=version,
     description="Rev.ai makes speech applications easy to build!",
     long_description=readme + '\n\n' + history,
-    author="Joshua Dong",
-    author_email='jdong42@gmail.com',
+    author="Kyle Bridburg",
+    author_email='kyle@rev.com',
```
amikofalvy: Should create an email alias? Example: revai-devs at rev.com.

We have a few existing aliases that we can use.

setup.py
```diff
@@ -62,5 +62,5 @@
     ],
     test_suite='tests',
     tests_require=test_requirements,
-    url='https://github.com/revdotcom/rev_ai',
+    url='https://github.com/revdotcom/revai-python-sdk',
```
amikofalvy: Are we still going to use the "SDK" suffix for the name? As per Joshes earlier comment, this usually signifies there are a large number of compiled binaries or C code in this repo. We don't have any of those at the moment.

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,139 @@
+# -*- coding: utf-8 -*-
+"""Speech recognition tools for using rev.ai
+"""
```
amikofalvy: minor -  strange line break please remove

src/rev_ai/__init__.py
amikofalvy: ask @JDongian for confirmation.

setup.py
amikofalvy: I'll ping you separately on slack about this.

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,139 @@
...
+            }
+            response = self.s.post(url_jobs, files=files)
+
+        return Job.from_json(response.json())
+
+    def get_job_details(self, id_):
+        """View information about a specific job.
+        The server will respond with the status and creation date.
+
+        :param id_: id that the server gave you
```
amikofalvy: We should clarify that this is the job id.

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,139 @@
...
+        url_jobs_id = urljoin(self.base_url, "jobs/{id_}".format(id_=id_))
+
+        response = self.s.get(url_jobs_id)
+
+        return Job.from_json(response.json())
+
+    def get_transcript_text(self, id_):
+        """Get the transcript of a specific job as json.
+
+        :param id_: id that the server gave you
```
amikofalvy: ditto - job id

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+        response = self.s.get(
+            url_jobs_transcript,
+            headers={'Accept': "text/plain"}
+        )
+
+        return response.text
+
+    def get_transcript_object(self, id_, format_):
```
amikofalvy: `format` param is not used. Please remove.

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+
+        url_jobs_transcript = urljoin(
+            self.base_url,
+            "jobs/{id_}/transcript".format(id_=id_)
+        )
+
+        response = self.s.get(
+            url_jobs_transcript,
+            headers={'Accept': 'application/{version}+json'
```
amikofalvy: Is there a naming collision with the version defined above?
https://github.com/revdotcom/revai-python-sdk/pull/3/files#diff-e8a9624e829d294d93268cd2b25fb8f2R20

We should rename to clarify that this is the transcript version and not the api version.

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,139 @@
...
+                .format(version="vnd.rev.transcript.v1.0")}
+        )
+
+        return Transcript.from_json(response.json())
+
+    def get_account(self):
+        """Get account information, such as remaining balance.
+        """
```
amikofalvy: strange split line

src/rev_ai/models/job_options.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+"""Job Submit Options Model"""
+
+
+class JobSubmitOptions:
+    def __init__(self,  metadata=None, callback_url=None, skip_diarization=None):
+        """
+        :param metadata: info to associate with the transcription job
+        :param callback_url: callback url to invoke on job completion as a webhook
```
amikofalvy: missing docs for `skip_diaraization`

src/rev_ai/apiclient.py
JDongian: this can be normal for python, I'm a promoter of this style but I don't enforce it on anyone

src/rev_ai/apiclient.py
```diff
@@ -0,0 +1,139 @@
...
+        :returns: raw response data
+        """
+        url_jobs = urljoin(self.base_url, "jobs")
+        payload = {
+            'media_url': media_url,
+            'metadata': options.metadata
+        }
+        if options.callback_url:
+            payload['callback_url'] = options.callback_url
+        response = self.s.post(url_jobs, json=payload)
```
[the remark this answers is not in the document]
JDongian: Make sure to put it in the line with `self.s.headers.update` so it applies to every request

src/rev_ai/apiclient.py
JDongian: no there isn't, that one is namespaced under `s`, via `s.version`. `version` here is the name of the parameter to `str.format`.

but I could see how this looks odd

src/rev_ai/streaming.py
```diff
@@ -84,4 +84,4 @@ def received_message(self, m):
     # c = RevClient("ws://localhost:8080/client/ws/speech")
```
JDongian: Consider deleting this file and restoring it later when the service is turned on

src/rev_ai/models/job_options.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+"""Job Submit Options Model"""
+
+
+class JobSubmitOptions:
+    def __init__(self,  metadata=None, callback_url=None, skip_diarization=None):
```
JDongian: I think `skip_diarization=False` is better. we're not going to use type annotations, but it shows that it is a boolean from the function sig

PR #4, opened 2019-01-18
AISDK-3: Add Jenkins file for deployment

kbridbur:


Jenkinsfile
```diff
@@ -0,0 +1,74 @@
...
+                '''
+            }
+        }
+        stage('Test') {
+            steps {
+                echo 'Testing..'
+                sh "python setup.py test"
+            }
+        }
+        stage("Version Check"){
```
kennygea: missing space after `(`

Jenkinsfile
```diff
@@ -0,0 +1,74 @@
...
+                    . ./sdk-deploy/bin/activate
+                    python setup.py sdist
+                    python setup.py bdist_wheel
+                '''
+            }
+        }
+        stage('Test') {
+            steps {
+                echo 'Testing..'
+                sh "python setup.py test"
```
kennygea: should this be run in the virtual env as well?

Jenkinsfile
```diff
@@ -0,0 +1,74 @@
+pipeline {
+    agent {
+        node 'linux-ci'
```
kennygea: minor, instead of straight up specifying `linux-ci`, it's better to use the node tags, so in this case it would be
```
agent {
    label 'linux && python'
}
```

note: I'm not 100% certain this is the exactly format but we should be able to specify labels.

see https://jenkins.io/doc/book/pipeline/jenkinsfile/#using-multiple-agents

kennygea: labels are given here: https://ci.rev.com/computer/linux-ci/

Jenkinsfile
```diff
@@ -0,0 +1,74 @@
...
+                    . ./sdk-deploy/bin/activate
+                    python -m pip install twine
+                    twine upload dist/* -u ${PYPI_USER_ID} -p ${PYPI_PASSWORD}
+                '''
+            }
+        }
+    }
+}
+
+def discoverVersion() {
```
kennygea: some comments like the xml docs we have in c# would be nice

Jenkinsfile
```diff
@@ -0,0 +1,117 @@
+pipeline {
+    agent {
+        label 'linux && python'
+    }
+    stages {
+        stage('Build') {
+            steps {
+                echo 'Building..'
+                sh '''
+                    virtualenv ./sdk-deploy
```
JDongian: specify specific python version. Consider using tox to set up the testing matrix.

Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+                    python setup.py bdist_wheel
+                '''
+            }
+        }
+        stage('Test') {
+            steps {
+                echo 'Testing..'
+                sh '''
+                    . ./sdk-deploy/bin/activate
+                    python setup.py test
```
JDongian: if not using tox for multiple version testing, also consider using the makefile.

Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+                '''
+            }
+        }
+        stage("Version Check") {
+            steps {
+                checkVersion(discoverVersion())
```
eugenep-rev: should we check the version before other stages?

tox.ini
```diff
@@ -0,0 +1,5 @@
+[tox]
+envlist = py27,py36,py37
```
JDongian: https://github.com/revdotcom/revai-python-sdk/blob/master/.travis.yml Kevin implemented 3.4 and 3.5 as well

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu:18.04
+
+RUN useradd -u 1001 -ms /bin/bash dockeruser
+
+RUN set -x \
```
kbridbur: Tox requires all versions of python to be installed. This dockerfile is purely for tests

Jenkinsfile
```diff
@@ -0,0 +1,124 @@
...
+
+    return matchesHttpsPrefix.replaceFirst('')
+}
+
+// sends notifications on failed builds or when the build becomes stable again
+def notifyComplete(String buildStatus) {
+    String SUCCESS = 'SUCCESS'
+    String FAILURE = 'FAILURE'
+
+    return
```
kbridbur: Line was added for testing, will be removed before merging

Jenkinsfile
```diff
@@ -0,0 +1,124 @@
...
+                    . ./sdk-deploy/bin/activate
+                    python setup.py sdist
+                    python setup.py bdist_wheel
+                '''
+            }
+        }
+        stage('Test') {
+            agent {
+                dockerfile {
+                    label 'linux'
```
kennygea: minor, this needs to be more specific, maybe use `linux && python` here?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    && pythonVersions='python2.7 python3.4 python3.5 python3.6 python3.7' \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends software-properties-common \
+    && apt-add-repository ppa:/deadsnakes/ppa \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends $pythonVersions \
+    && apt-get purge -y --auto-remove software-properties-common \
```
kennygea: what is this purge and rm -rf commands doing?

Jenkinsfile
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Tag repo with version number so we know this version has been released
+def tagRepo(tag) {
+    echo "tagging repo with ${tag}"
+
+    // deletes current tag locally, this will make sure if the tag is deleted from remote it would be set again
+    sh "git tag -d ${tag} || (exit 0)"
+    // see if it's still on remote - needed to avoid overwriting the tag
+    sh "git fetch https://${env.GIT_CREDENTIALS}@${getRepoUrl()} --tags"
```
kennygea: what's the purpose of this line?

kbridbur: I'm actually not sure. The comment says it checks if the tag is still on remote but we don't do anything with taht information. These portions of the script are copied from revcommon

PR #7, opened 2019-01-23
AISDK-2: Add testing to Python SDK

k-weng:


src/rev_ai/models/__init__.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Models"""
 
-from models.job import Job
-from models.job_options import JobSubmitOptions
-from models.account import Account
-from models.transcript import Transcript
\ No newline at end of file
+from .job import Job
```
amikofalvy: @JDongian , is it best practice to use explicit relative imports?
https://stackoverflow.com/questions/22511792/python-from-dotpackage-import-syntax

JDongian: it's standard practice.

tests/test_account.py
```diff
@@ -0,0 +1,34 @@
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
+import pytest
```
JDongian: is there are reason you use pytest over unittest?
My only reason to prefer unittest is that it is built in.

k-weng: Not a particular personal reason, I chose it after looking at online opinions of `unittest` versus `pytest`. Personally, I don't have an attachment to either. It shouldn't be too hard to change from one to the other if you or anyone else has a strong preference towards `unittest` 🙂

kbridbur: I think it's fine to use `pytest`

tests/test_account.py
```diff
@@ -0,0 +1,34 @@
...
+        )
+
+    def test_get_account_not_authorized_error(self, mockclient):
+        data = {
+            "title": "Authorization has been denied for this request",
+            "status": 401
+        }
+        mockclient.session.get.return_value.json.return_value = data
+
+        with pytest.raises(KeyError):
```
kbridbur: These tests will change when AISDK-18 is implemented

src/rev_ai/models/account.py
```diff
@@ -11,10 +11,16 @@ def __init__(self, email, balance_seconds):
         self.email = email
         self.balance_seconds = balance_seconds
 
+    def __eq__(self, other):
```
kennygea: @k-weng. I thought we discussed this be we can extend it to a more general question.That is, do we actually want to override the default equality operator? I say no, since from a user perspective, isn't this not obvious that we do so.  So say I return two `account` objects from this SDK, as a developer, I would expect them to NOT have equality as an object (In c# the default is that this implies it's referencing the same pointer in memory, I'm assuming this is the same in python). I would remove these equality overrides and just create a helper for testing shallow equality between two objects.

JDongian: meh, this isn't a language with pointer manipulation. I think this good.

If you actually wanted to know if something points to the same thing in memory, in python, you would use `id`. But nobody does that.

requirements_dev.txt
```diff
@@ -0,0 +1 @@
+pytest
```
JDongian: can you pin specific versions for all the requirements? It's much easier to avoid a hassle in the future when breaking changes happen.

tests/test_account.py
JDongian: Okay, then we should just use one. There are unittest mock imports in addition to pytest imports.

tests/test_job.py
```diff
@@ -0,0 +1,150 @@
...
+        data = {
+            'id': JOB_ID,
+            'status': status,
+            'created_on': created_on,
+            'metadata': metadata
+        }
+        mockclient.session.post.return_value.json.return_value = data
+
+        res = mockclient.submit_job_url(MEDIA_URL,
+                                        JobSubmitOptions(metadata=metadata))
```
JDongian: To be pedantic, do you want to also test the callback_url functionality (found it was missing via coverage command)?

src/rev_ai/models/account.py
k-weng: @kennygea I originally had this override before our discussion and hadn't changed it yet, but now I think this is fine - `__eq__` or `==` should imply equality in attributes. In Python to check for reference equality, you would use `is`. In addition, the helper for assert would simply just be the same implementation as the overridden equality implementation.

tests/test_account.py
k-weng: Sounds good - for mocking, the `mock` package was added to be part of the `unittest` package but I realized the `pytest` devs made a pytest wrapper `pytest-mock` so I will convert them appropriately.

src/rev_ai/models/account.py
JDongian: right, 
```
In [147]: a = set([1,2])

In [148]: b = set([2,1])

In [149]: id(a), id(b)
Out[149]: (139984634142792, 139984634143688)

In [150]: a == b
Out[150]: True

In [151]: a, b
Out[151]: ({1, 2}, {1, 2})

In [152]: a is b
Out[152]: False
```

JDongian:
```
----------- coverage: platform linux, python 3.6.6-final-0 -----------
Name                               Stmts   Miss  Cover
------------------------------------------------------
src/__init__.py                        1      0   100%
src/rev_ai/__init__.py                 2      0   100%
src/rev_ai/apiclient.py               48      3    94%
src/rev_ai/models/__init__.py          6      0   100%
src/rev_ai/models/account.py          11      1    91%
src/rev_ai/models/job.py              21      1    95%
src/rev_ai/models/job_options.py      10      3    70%
src/rev_ai/models/job_status.py        8      0   100%
src/rev_ai/models/transcript.py       33      9    73%
------------------------------------------------------
TOTAL                                140     17    88%
```
Missing tests for:
* eq operators
* if not api_key: raise ValueError("API Key cannot be empty.")

Other than updating the Makefile for a working `make coverage` command (is it in scope?), this looks good to me.

kbridbur:
I don't think we need to uphold any specific coverage standards, I think keeping our testing standards the same as in other rev projects is acceptable. I would prefer that we add a test for "if not api_key" but I think checking all the equality operators isn't necessary. 

Looks really good otherwise. I would add that one test, fix `make coverage` and merge

tests/test_job.py
```diff
@@ -0,0 +1,159 @@
...
+JOB_ID = '1'
+MEDIA_URL = "https://example.com/test.mp3"
+CALLBACK_URL = 'https://callback.com/'
+
+
+@pytest.mark.usefixtures("mockclient")
+class TestJobEndpoints():
+    def test_get_job_details_success(self, mockclient):
+        status = 'transcribed'
+        created_on = '2018-05-05T23:23:22.29Z'
```
kbridbur: Is there a reason you didn't pull this parameter out into a global constant like JOB_ID? If not please do

Issue #8, opened 2019-01-24
Backend also updated?


kalling:
Since you guys did the API cleanup I have not been able to submit a job with my old code using 0.0.3

It fails during json decoding.  (It expecting something, but there was nothing)

I also tried cloning your current version.  (Even though that's not what you have on pypi)
and after fixing a formatting issue in apiclient.py I did a submission, but it fails in the json decoder just like it does with the 0.0.3 version.

BTW, I think that submit_job_local_file() and submit_job_url() would be better if you kept options as optional with sensible defaults instead of forcing the construction of an options class.

kbridbur:
Hi Richard, sorry to hear about your problem. I'm happy to help you debug the problem, it would be very helpful if you could send me the error and a snippet of the code that's causing it so I can reproduce. If you could send me those at kyle@rev.com I'll take a look.

Also, I completely agree with your suggestion for the options parameter, I'll add that change in my next PR

kbridbur:
Posting here to say that this issue should be solved. The API previously had an issue with very long files, this should be fixed now

PR #9, opened 2019-01-26
 AISDK-18: Better error handling for SDK

k-weng:


src/rev_ai/apiclient.py
```diff
@@ -28,51 +28,79 @@ def __init__(self, api_key):
...
         """
-        url_jobs = urljoin(self.base_url, "jobs")
-        payload = {
-            'media_url': media_url,
-            'metadata': options.metadata
-        }
-        if options.callback_url:
-            payload['callback_url'] = options.callback_url
+        if not media_url:
+            raise ValueError('media_url must be provided')
```
JDongian: this seems extra, why would the user provide a null argument to submit_job_url?

I don't think APIs should have excess safety logic. IMO raw as-close-to-the-metal is preferrable

setup.py
```diff
@@ -36,17 +36,18 @@
 setup(
     name='rev_ai',
     version=version,
-    description="Rev.ai makes speech applications easy to build!",
+    description='Rev.ai makes speech applications easy to build!',
```
JDongian: Why is it necessary to convert all quotes to single quotes?

If you are establishing a coding style, I'm curious where it is documented. Speech team's coding style is outlined in https://speech-knowledge.rev.com/intro/coding-conventions (`flake8 --ignore=E122,E126,E127,E501`)

src/rev_ai/models/__init__.py
```diff
@@ -2,7 +2,6 @@
 """Models"""
 
 from .job import Job
-from .job_options import JobSubmitOptions
 from .job_status import JobStatus
```
JDongian: off topic note: consider adding this to src/rev_ai/__init__.py , it looks like the end user may end up using it and having a convenient import is helpful

setup.py
k-weng: Thought it was better to be consistent, and my preference was towards single quotes. As for documentation of coding styles - we don't have documentation of Python's but probably a good idea - will most likely be on `dev-wiki` if we do create one

src/rev_ai/apiclient.py
```diff
@@ -28,51 +28,79 @@ def __init__(self, api_key):
...
+
         self.session = requests.Session()
         self.session.headers.update({
-            'Authorization': 'Bearer {api_key}'.format(api_key=api_key),
+            'Authorization': 'Bearer {}'.format(api_key),
             'User-Agent': 'python_sdk'
         })
 
-    def submit_job_url(self, media_url, options):
+    def submit_job_url(
```
k-weng: Per suggestion and comments from the channel and issue https://github.com/revdotcom/revai-python-sdk/issues/8, I removed the `JobSubmitOptions` and opted for simply adding optional parameters to both of the `submit_job_...` methods.

src/rev_ai/apiclient.py
JDongian: I think it's a bit arbitrary to just check falesy values, you could go further and also validate URL schema and stuff like that. But I can see your point.

src/rev_ai/apiclient.py
```diff
@@ -28,51 +28,79 @@ def __init__(self, api_key):
...
 
         with open(filename, 'rb') as f:
             files = {
                 'media': (filename, f),
                 'options': (None, json.dumps(payload))
             }
-            response = self.session.post(url_jobs, files=files)
+
+            response = self.session.post(url, files=files)
+            response.raise_for_status()
```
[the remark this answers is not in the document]
JDongian: Yeah this is probably a lot easier for both users and us. Good find.

JDongian:
also... can we please upload to https://pypi.org/project/rev-ai/ ?

I suggest that we update the issue template to have
- [ ] updated version number
- [ ] uploaded new version to pypi (after PR merge)

what do you think? or should we use github tags instead/alongside?

k-weng:
I don't exactly know the release / publish process that was set for the SDKs - @kbridbur would know.

kbridbur:
I'm still working on the deployment scripts, having some trouble getting tox to work. Once that works we will have a ci job anyone can run to release a new version of the SDK based on the master branch of the repo

src/rev_ai/apiclient.py
kbridbur: My impression of what was wanted in that discussion is just options=None as the default. Not total removal of the object. I think we should revert this change. What do you think?

k-weng: I think in general, I find it strange to have optional arguments encapsulated into a separate class in Python. I think it's much cleaner and simpler for the user to have each option as an optional parameter of the methods themselves. I might have missed the discussion about this though.

kbridbur: I think that makes sense. I know very little about python conventions. If @JDongian agrees I think we can keep it this way

src/rev_ai/apiclient.py
```diff
@@ -28,51 +28,79 @@ def __init__(self, api_key):
...
         """Submit media given a URL for transcription.
         The audio data is downloaded from the URL.
 
         :param media_url: web location of the media file
-        :param options: JobSubmitOptions object for the job
+        :param metadata: info to associate with the transcription job
+        :param callback_url: callback url to invoke on job completion as a
+                             webhook
+        :param skip_diarization: should rev.ai skip diaization when
+                                 transcribing this file
```
alexsku: i'm curious what is the acceptable line length for python, like if it was c# i would say the lines above should not be wrapped

src/rev_ai/models/transcript.py
```diff
@@ -36,16 +36,18 @@ def __init__(self, speaker, elements):
...
     @classmethod
     def from_json(cls, json):
         """Alternate constructor used for parsing json"""
         return cls(
-            json["speaker"],
-            [Element.from_json(element) for element in json.get("elements", [])]
-        )
+            json['speaker'],
+            [Element.from_json(element)
+             for element in json.get('elements', [])])
```
alexsku: just curious why is this line wrapped

src/rev_ai/apiclient.py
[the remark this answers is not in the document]
alexsku: should we update methods documentation explaining that httperror might be thrown and maybe somewhere describe that thhe response would carry problem details?

src/rev_ai/apiclient.py
kbridbur: Acceptable line length is 80 characters

alexsku: just curious why 80? where is it coming from?

kbridbur: I'm not sure. Maybe @JDongian can give a more indepth reason but it is the standard
https://www.python.org/dev/peps/pep-0008/

k-weng: I think we can set it to be longer if needed - I think most change it to like 100 or 120 or just plain ignore it

src/rev_ai/models/transcript.py
k-weng: It hits the 80 character limit.

kennygea: 80 character limit does seem really crazy in this case

k-weng: Will be set to 100

PR #11, opened 2019-02-01
AISDK-22: Add delete method to SDK client

k-weng:


src/rev_ai/apiclient.py
```diff
@@ -159,6 +159,24 @@ def get_transcript_object(self, id_):
...
+        :raises: HTTPError
+        """
+        if not id_:
+            raise ValueError('id_ must be provided')
+
+        url = urljoin(self.base_url, 'jobs/{}'.format(id_))
+        response = self.session.delete(url)
+        response.raise_for_status()
+
+        return True
```
dkokotov: why not just `return`? It seems that this can never return anything other than true - if delete fails, it will throw - so there doesn't seem to be a reason to return boolean.

PR #12, opened 2019-02-04
AISDK-23: Change base url for RevAi API

k-weng:


kbridbur:
@JDongian I don't think these changes are breaking. Old versions on the SDK will still function

JDongian:
Ok so it's probably a minor bump?

PR #13, opened 2019-02-04
AISDK-26: Include version string in user agent

kbridbur:


src/rev_ai/apiclient.py
```diff
@@ -98,9 +98,14 @@ def submit_job_local_file(
         if callback_url:
             payload['callback_url'] = callback_url
 
+        illegal_chars = ['\\','~','/']
```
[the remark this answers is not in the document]
kbridbur: I don't think we should. This is purely for filename upload and I don't think that is necessary or helpful from a user standpoint

src/rev_ai/apiclient.py
```diff
@@ -3,6 +3,8 @@
 
 import requests
 import json
+import re
```
JDongian: This isn't necessary. You should be able to just `from . import __version__`.

src/rev_ai/apiclient.py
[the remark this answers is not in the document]
JDongian: I can agree with that. Expanding path to abspath is still nice tho

tparnell:
You may also want to screen for bidirectional characters.  S3 can deal with them, but the .NET SDK cannot.  See FOX-20165 and search for `bidi` in Jira.

kbridbur:
Thanks @tparnell. I'm reducing the scope of this issue because I think handling this should be the responsibility of the API itself. I'll be filing an issue to handle this on our end

tests/test_apiclient.py
```diff
@@ -12,7 +13,7 @@ def test_constructor_with_success(self):
         client = RevAiAPIClient(token)
 
         headers = client.session.headers
-        assert headers.get('User-Agent') == 'python_sdk'
+        assert headers.get('User-Agent') == 'python_sdk-{}'.format(__version__)
```
amikofalvy: I think this test should use string literals.
```suggestion
        # Setup
        __version__ = 'test-version'
        
        # Assert
        assert headers.get('User-Agent') == 'python_sdk-test-version'
```

amikofalvy: Currently, the test just repeats the implementation code.

src/rev_ai/apiclient.py
```diff
@@ -40,7 +41,7 @@ def __init__(self, access_token):
         self.session = requests.Session()
         self.session.headers.update({
             'Authorization': 'Bearer {}'.format(access_token),
-            'User-Agent': 'python_sdk'
+            'User-Agent': 'python_sdk-{}'.format(__version__)
```
alexsku: are f strings available for us here?

kbridbur: No, f strings were introduced in python 3.6 to my knowledge.

alexsku: i was afraid so :(

tests/test_apiclient.py
alexsku: @amikofalvy do you mean we should check that the user-agent string is `'python_sdk-{}'.format(__version__)` versus `python_sdk-test-version'`? I personally would have put it as Kyle did it i think but i don't have strong opinion on this

kbridbur: Talked with @amikofalvy. We ended up agreeing on this implementation

PR #14, opened 2019-02-05
AISDK-27: Add get_transcript_json method

kbridbur:


src/rev_ai/apiclient.py
```diff
@@ -142,6 +142,23 @@ def get_transcript_text(self, id_):
...
+        :param id_: id of job to be requested
+        :returns: transcript data as json
+        :raises: HTTPError
+        """
+        if not id_:
+            raise ValueError('id_ must be provided')
+
+        url = urljoin(self.base_url, 'jobs/{}/transcript'.format(id_))
+        response = self.session.get(
+            url, headers={'Accept': 'application/{}+json'.format('vnd.rev.transcript.v1.0')})
```
JDongian: do you want to store the value `vnd.rev.transcript.v1.0` in a class variable

kbridbur: Seems reasonable

src/rev_ai/apiclient.py
```diff
@@ -142,6 +142,23 @@ def get_transcript_text(self, id_):
...
+        response.raise_for_status()
+
+        return response.json()
+
     def get_transcript_object(self, id_):
         """Get the transcript of a specific job as json.
```
k-weng: The docstring for `get_transcript_object` is the same as `get_transcript_json`

PR #20, opened 2019-03-05
AISDK-21: Add Get Jobs endpoint to Python SDK

k-weng:


Jenkinsfile
```diff
@@ -106,12 +106,12 @@ def notifyComplete(String buildStatus) {
...
     }
-    
+
     // build status of null means successful
     buildStatus =  buildStatus ?: SUCCESS
     def prevBuildResult = currentBuild.getPreviousBuild()?.getResult()
     echo "notifyComplete with status ${buildStatus}, previous build was ${prevBuildResult}"
-        
+
     if (buildStatus == FAILURE)
```
kbridbur: Please change this as per the notes on the issue
```suggestion
    if (buildStatus == FAILURE && prevBuildResult != FAILURE)
```

Jenkinsfile
```diff
@@ -120,6 +120,6 @@ def notifyComplete(String buildStatus) {
     else if (buildStatus == SUCCESS && prevBuildResult != SUCCESS)
         slackSend (
             color: '#00FF00', // green
-            channel: '#rev-ai-dev', 
+            channel: '#rev-ai-dev',
```
kbridbur: Please change this as per the notes on the issue
```suggestion
            channel: '#revai-alerts-nonprod',
```

kbridbur: Do the same with the failure message

README.md
```diff
@@ -65,9 +65,37 @@ job_details = client.get_job_details(job.id)
...
+You can retrieve a list of transcription jobs with optional parameters
+
+```python
+jobs = client.get_list_of_jobs()
+
+# limit amount of retrieved jobs
+jobs = client.get_list_of_jobs(limits=3)
+
+# get jobs starting after a certain job id
+jobs = client.get_list_of_jobs(starting_after='3')
```
kbridbur: I think '3' is a bad example here since it implies this argument is a number of jobs to start after as opposed to what it actually is: a jobid to start after. I would either add a comment or change the value to something like `another_job.id`

Jenkinsfile
amikofalvy: Please use channelID (`CFPMB0BK4`) instead of channel name. Make sure to leave a comment describing that that the channel Id refers to #revai-alerts-nonprod.

ChannelId does not change when we rename the slack channel

README.md
amikofalvy: Or you can even use a real jobId as an example.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.2.1
+current_version = 2.3.0
```
amikofalvy: This should match the version in the other issue AISDK-29

PR #21, opened 2019-03-05
AISDK-29: Change User-Agent for Python SDK

k-weng:


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.2.1
+current_version = 2.3.0
```
kbridbur: I think 2.2.2 is fine. I would consider this a patch version since its backwards compatible and has no visible effects on the SDK from a user perspective

PR #25, opened 2019-06-14
AISDK-37: Add basic RevAiStreamingClient to the python SDK

Ymardini:


src/streaming/streamingclient.py
```diff
@@ -0,0 +1,178 @@
...
+        self.on_close = on_close
+        self.on_open = on_open
+        self.thread = SpeechThread
+        self.client = websocket.WebSocket(enable_multithread = True, 
+            on_error = self.on_error, 
+            on_close = self.on_close, 
+            on_open = self.on_open
+            )
+
+    def start(self):
```
kbridbur: This method should return the responses() generator and take a generator as input.

My reasoning for this is that I don't believe a customer would want to start the streaming connection and not begin sending and receiving from the server. By doing this we can combine 3 method calls into one. 

Before
```
client.start()
client.send_data_as_gen(input)
for i in client.responses(): ...
```
After
```
responses = client.start(input)
for i in responses: ...
```

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if hasattr(self, 'request_thread'):
+            if self.request_thread.isAlive():
+                raise ValueError("""Data is still being sent and will interfere 
+                    with the responses.""")
+
+        self.request_thread = self.thread(self.RequestSending, [generator])
+        self.request_thread.start()
+
+    def send_data_as_filepath(self, filepath):
```
kbridbur: I don't believe that this method fit in this class. The responsibility of this class is to send and receive data from rev.ai. This filepath method gives the client more responsibility and makes it harder for us in the future. 

For example, currently if we're passed a file which we don't have permission to read the read operation will error and it will look like the client is erroring. In general we should avoid interacting with files as it puts the responsibility of handling a lot of edge cases on us.

Customers expect that if we say we're going to take responsibility for something (in this case turning their file into a generator) that we will gracefully handle any errors that occur during that process. This method currently doesn't do that and actually handling it way more work than we want to do

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,178 @@
...
+            on_close = self.on_close, 
+            on_open = self.on_open
+            )
+
+    def start(self):
+        """Function to connect the websocket to the URL and start the response thread
+        """
+        url = self.base_url + '?access_token={}'.format(self.access_token) \
+              + '&content_type={}'.format(self.config.get_content_type_string()
+                )
```
kbridbur: Minor formatting: This `)` should be on the previous line

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,178 @@
...
+            raise ValueError('filepath must be provided')
+
+        with io.open(filepath, 'rb') as stream:
+            while True:
+                piece = stream.read(8192)
+                if not piece:
+                    break
+                yield piece
+
+    def RequestSending(self, generator):
```
kbridbur: Formatting: The python convention for private methods is snake case starting with an underscore (ie `_request_sending`)

kbridbur: I would also rename this method to be more clear. Something like `send_data`

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,178 @@
...
+            elif opcode == websocket.ABNF.OPCODE_CLOSE:
+                if response and len(response) >= 2:
+                    code = 256 * six.byte2int(response[0:1]) + six.byte2int(response[1:2])
+                    reason = response[2:].decode('utf-8')
+                    self.on_close(code, reason)
+                return
+            else:
+                yield ''
+
+class SpeechThread(threading.Thread):
```
kbridbur: It seems like you're trying to make this class generic when in reality when only use this for one purpose. I would simplify this by defining it in the class as a separate thread with only the responsibility of sending data from a generator.

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,178 @@
...
+            self.client.send_binary(chunk)
+
+        self.client.send("EOS")
+
+    def responses(self):
+        """A generator of reponses from the server. Yields the data decoded.
+        """
+        while True:
+            with self.client.readlock:
+                opcode, response = self.client.recv_data()
```
kbridbur: I don't think response is a very fitting name for this variable. The response from the server consists of two pieces (when considering only the pieces we want to use). The opcode of the response and the data within the response.
We should follow the convention the websocket package uses and call these `opcode` and `data`

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,178 @@
...
+    pass
+
+class RevAiStreamingClient():
+    def __init__(self, 
+                access_token, 
+                config, 
+                version = 'v1alpha',  
+                on_error = on_error, 
+                on_close = on_close, 
+                on_open = on_open):
```
kbridbur: We should replace this with a function for `on_connected` and call `on_connected` when we receive a connected message from the server. I think this better serves the purpose a user would want, the connected message is a better signifier of successfully beginning a connection with our service AND provides them with their job id

src/streaming/AudioConfig.py
```diff
@@ -0,0 +1,29 @@
...
+        self.channels = channels
+
+    def get_content_type_string(self):
+        """Returns the content type and params as a string for the websocket
+           connection
+        """
+        return self.content_type +  \
+            (';layout={}'.format(self.layout) if self.layout else '') + \
+            (';rate={}'.format(self.rate) if self.rate else '') + \
+            (';format={}'.format(self.format) if self.format else '')+ \
```
kbridbur: space before `+`

src/streaming/AudioConfig.py
```diff
@@ -0,0 +1,29 @@
...
+        self.layout = layout
+        self.rate = rate
+        self.format = audio_format
+        self.channels = channels
+
+    def get_content_type_string(self):
+        """Returns the content type and params as a string for the websocket
+           connection
+        """
+        return self.content_type +  \
```
kbridbur: Extra space after the `+`

src/streaming/AudioConfig.py
```diff
@@ -0,0 +1,29 @@
...
+        """
+        self. content_type = content_type
+        self.layout = layout
+        self.rate = rate
+        self.format = audio_format
+        self.channels = channels
+
+    def get_content_type_string(self):
+        """Returns the content type and params as a string for the websocket
+           connection
```
kbridbur: This can probably go on the same line

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+import io
+import threading
+import six
+import json
+
+def on_error(error):
+    raise error
+
+def on_close(code, reason):
+    print("{} : {}".format(code, reason))
```
kbridbur: We should add a short message here like for connected. Something like `Connection Closed with {code}: {reason}`

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        if not generator:
+            raise ValueError('generator must be provided')
+
+        if hasattr(self, 'request_thread'):
+            if self.request_thread.isAlive():
+                raise ValueError("""Data is still being sent and will interfere with the responses.""")
+
+        self.request_thread = threading.Thread(target = self._send_data, 
+                                args = [generator])
```
kbridbur: This can be on the previous line

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        if not generator:
+            raise ValueError('generator must be provided')
+
+        for chunk in generator:
+            self.client.send_binary(chunk)
+
+        self.client.send("EOS")
+
+    def _responses(self):
```
kbridbur: I would give this a slightly more descriptive name. Something like `_get_responses`

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+                opcode, data = self.client.recv_data()
+            if six.PY3 and opcode == websocket.ABNF.OPCODE_TEXT:
+                dec_data = data.decode('utf-8') 
+                data_dict = json.loads(dec_data)
+                if data_dict['type'] == 'connected':
+                    yield self.on_connected(data_dict['id'])
+                else:
+                    yield dec_data
+            elif opcode == websocket.ABNF.OPCODE_TEXT:
+                data_dict = json.loads(dec_data)
```
kbridbur: I don't think `dec_data` exists here.

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _responses(self):
+        """A generator of reponses from the server. Yields the data decoded.
+        """
+        while True:
+            with self.client.readlock:
+                opcode, data = self.client.recv_data()
+            if six.PY3 and opcode == websocket.ABNF.OPCODE_TEXT:
+                dec_data = data.decode('utf-8') 
+                data_dict = json.loads(dec_data)
```
kbridbur: Lets combine this and the previous line into one

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+                if data_dict['type'] == 'connected':
+                    yield self.on_connected(data_dict['id'])
```
kbridbur: I think yielding this will either error since we're yeilding a void or will yield `None`. Instead we should call `self.on_connected` then pass. Python generators continue to run until a `yield` or `return` is encountered so if we just pass here the generator will keep looking at the responses until we get to a hypothesis message

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+                else:
+                    yield dec_data
+            elif opcode == websocket.ABNF.OPCODE_TEXT:
+                data_dict = json.loads(dec_data)
+                if data_dict['type'] == 'connected':
+                    yield self.on_connected(data_dict['id'])
```
kbridbur: Same as above https://github.com/revdotcom/revai-python-sdk/pull/25/files#r294533377

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,132 @@
...
+        self.access_token = access_token
+        self.config = config
+        self.base_url = base_url = 'wss://api.rev.ai/speechtotext/{}/stream'.format(version)
+        self.on_error = on_error
+        self.on_close = on_close
+        self.on_connected = on_connected
+        self.client = websocket.WebSocket(enable_multithread = True, 
+            on_error = self.on_error, 
+            on_close = self.on_close, 
+            on_connected = self.on_connected)
```
kbridbur: `)` should go on next line. As a general rule we try to separate parenthesis in one of two ways
```
(
    stuff
)
```
or 
`(stuff)`

src/streaming/streamingclient.py
Ymardini: Do you mean combine as in data_dict = json.loads(data.decode('utf-8'))? I have it as two because I need to yield dec_data in the else case.

src/streaming/AudioConfig.py
```diff
@@ -0,0 +1,29 @@
+"""
+    Configuration for an audio stream
+"""
+
+class AudioConfig():
```
amikofalvy: If we support Audio and Video I think naming this audioconfig is misleading. Perhaps we should rename it MediaConfig

src/streaming/AudioConfig.py
amikofalvy: @kbridbur  There are python linters. Do we use them?

JDongian: `flake8 --ignore=E122,E126,E127,E501`

JDongian: https://speech-knowledge.rev.com/intro/coding-conventions

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        if not access_token:
+            raise ValueError('access_token must be provided')
+
+        if not config:
+            raise ValueError('config must be provided')
+
+        self.access_token = access_token
+        self.config = config
+        self.base_url = base_url = 'wss://api.rev.ai/speechtotext/{}/stream'. \
```
michellekhuang: consider using f-strings if on python3.6+ it makes formatting a lot easier. https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498

src/streaming/AudioConfig.py
kbridbur: There are and I believe we use flake8. I'll double check

src/streaming/streamingclient.py
kbridbur: We want to support python 2.7 and 3.4+ so I don't think we can use these

tests/test_streamingclient.py
```diff
@@ -0,0 +1,94 @@
...
+        url = mock_streaming_client.base_url + \
+            '?access_token={}'.format(mock_streaming_client.access_token) + \
+            '&content_type={}'. \
+            format(mock_streaming_client.config.get_content_type_string())
+        if six.PY3:
+            example_data = '{"type":"partial","transcript":"Test"}'.encode('utf-8')
+        else:
+            example_data = '{"type":"partial","transcript":"Test"}'
+        data = [
+            [websocket.ABNF.OPCODE_TEXT, '{"type":"connected","id":"testid"}'.encode('utf-8')],
```
kbridbur: I think you also need to change this message based on python version `six.PY3`

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,144 @@
+"""StreamingClient tool used for streaming services"""
+from src.streaming.MediaConfig import MediaConfig
+import websocket
```
kbridbur: You'll need to add this to requirements.txt

amikofalvy: do you mean the `__index__.py` file ?

amikofalvy: I think the MediaConfig file did not get added to git

kbridbur: We don't use an `__index__.py` file. pip only requires a `requirements.txt` file with requirements and their versions

src/streaming/AudioConfig.py
kbridbur: IT seems like this class is still here. Make sure you remove it

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,144 @@
...
+        """Constructor for Streaming Client
+
+        :param access_token: access token which authorizes all requests and 
+            links them to your account. Generated on the settings page of your 
+            account dashboard on Rev.ai.
+        :param config: an MediaConfig object containing audio information. 
+            See MediaConfig.py for more information
+        :param version (optional): version of the streaming speech to text as  
+            string
+        :param on_message (optional): function to be called when recieving a
```
kbridbur: This is no longer a param. Please remove this

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,144 @@
...
+                on_close = on_close, 
+                on_connected = on_connected):
+        """Constructor for Streaming Client
+
+        :param access_token: access token which authorizes all requests and 
+            links them to your account. Generated on the settings page of your 
+            account dashboard on Rev.ai.
+        :param config: an MediaConfig object containing audio information. 
+            See MediaConfig.py for more information
+        :param version (optional): version of the streaming speech to text as
```
kbridbur: Lets make this a little more clear. `version of the streaming api to be used`

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,144 @@
...
+                version = 'v1alpha',  
+                on_error = on_error, 
+                on_close = on_close, 
+                on_connected = on_connected):
+        """Constructor for Streaming Client
+
+        :param access_token: access token which authorizes all requests and 
+            links them to your account. Generated on the settings page of your 
+            account dashboard on Rev.ai.
+        :param config: an MediaConfig object containing audio information.
```
kbridbur: `a MediaConfig`

src/streaming/streamingclient.py
```diff
@@ -0,0 +1,144 @@
...
+        self.access_token = access_token
+        self.config = config
+        self.base_url = base_url = 'wss://api.rev.ai/speechtotext/{}/stream'. \
+            format(version)
+        self.on_error = on_error
+        self.on_close = on_close
+        self.on_connected = on_connected
+        self.client = websocket.WebSocket(
+            enable_multithread = True, 
+            on_error = self.on_error,
```
kbridbur: https://github.com/websocket-client/websocket-client/blob/master/websocket/_core.py

It looks like `on_error`, `on_close` and `on_connected` do not exist in the websocket constructor. Please remove these lines

tests/test_streamingclient.py
```diff
@@ -0,0 +1,96 @@
...
+        example_client = RevAiStreamingClient(example_token, example_config)
+
+        assert example_client.access_token == 'token'
+        assert example_client.config == example_config
+
+    def test_constructor_no_token_no_config(self):
+        example_token = 'token'
+        example_config = MediaConfig()
+
+        with pytest.raises(Exception):
```
JDongian: you want the specific exception

tests/test_streamingclient.py
```diff
@@ -0,0 +1,96 @@
...
+            RevAiStreamingClient(config = example_config)
+
+    def test_start_success(self, mock_streaming_client, mock_generator, capsys):
+        url = mock_streaming_client.base_url + \
+            '?access_token={}'.format(mock_streaming_client.access_token) + \
+            '&content_type={}'. \
+            format(mock_streaming_client.config.get_content_type_string())
+        if six.PY3:
+            example_data = '{"type":"partial","transcript":"Test"}'.encode('utf-8')
```
JDongian: consider a way to avoid duplication of constants

tests/test_streamingclient.py
```diff
@@ -0,0 +1,96 @@
...
+        mock_streaming_client.client.send_binary.assert_any_call(1)
+        mock_streaming_client.client.send_binary.assert_any_call(2)
+        assert hasattr(mock_streaming_client, 'request_thread')
+        for ind, response in enumerate(response_gen):
+            assert capsys.readouterr().out == exp_responses[ind]
+            assert exp_responses[ind+1] == response
+        assert capsys.readouterr().out == exp_responses[2]
+
+    def test_start_failure_to_connect(self, mock_streaming_client, mock_generator):
+        mock_streaming_client.client.connect = lambda x : 1/0
```
JDongian: Seems fine, but I would have gone for something like
```
lambda _: raise RuntimeError
```

src/rev_ai/streamingclient.py
```diff
@@ -0,0 +1,139 @@
...
+
+        :param generator: generator object that yields binary audio data
+        """
+        url = self.base_url + '?access_token={}'.format(self.access_token) \
+            + '&content_type={}'.format(self.config.get_content_type_string())
+        try:
+            self.client.connect(url)
+            self._start_send_data_thread(generator)
+            return self._get_response_generator()
+        except Exception as e:
```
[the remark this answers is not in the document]
JDongian: oh right because you can override the handler, derp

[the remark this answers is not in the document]
JDongian: derp, it's because you want to be able to override the behavior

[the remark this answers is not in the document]
Ymardini: This provides the user the ability to have a custom error function depending on the error raised by the websocket connection. The default is set to raise. I have it as a generic exception because I'm sure there are multiple possible errors to be obtained via initializing the websocket connection.

PR #26, opened 2019-06-21
AISDK-51: Add streaming responses to SDK

kbridbur:
Todo:
rebase

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ captions = client.get_captions(job.id)
 
 ### Streamed outputs
 
-Any output format can be retrieved as a stream for convenience. In these cases we return the raw http response to you. The output can be retrieved via `response.content`, `response.iter_lines()` or `response.iter_content()`.
+Any output format can be retrieved as a stream. In these cases we return the raw http response to you. The output can be retrieved via `response.content`, `response.iter_lines()` or `response.iter_content()`.
 
 ```python
 text_stream = client.get_transcript_text_as_stream(job.id)
--- a/README.md
+++ b/README.md
@@ -113,7 +113,27 @@ of the [Get Transcript](https://www.rev.ai/docs#operation/GetTranscriptById) end
 when using the json response schema. While the text output is a string containing
 just the text of your transcript
 
-### Streaming audio
+### Getting captions output
+
+You can also get captions output from the SDK:
+
+```python
+captions = client.get_captions(job.id)
+```
+
+### Streamed outputs
+
+Any output format can be retrieved as a stream for convenience. In these cases we return the raw http response to you. The output can be retrieved via `response.content`, `response.iter_lines()` or `response.iter_content()`.
+
+```python
+text_stream = client.get_transcript_text_as_stream(job.id)
+
+json_stream = client.get_transcript_json_as_stream(job.id)
+
+captions_stream = client.get_captions_as_stream(job.id)
+```
+
+## Streaming audio
 
 In order to stream audio, you will need to setup a streaming client and a media configuration for the audio you will be sending.
 
--- a/src/rev_ai/apiclient.py
+++ b/src/rev_ai/apiclient.py
@@ -179,12 +179,10 @@ class RevAiAPIClient:
 
         return [Job.from_json(job) for job in response.json()]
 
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -112,3 +112,23 @@ Both the json and object forms contain all the formation outlined in the respons
...
+
+You can also get captions output from the SDK:
+
+```python
+captions = client.get_captions(job.id)
+```
+
+## Streamed outputs
+
+Any output format can be retrieved as a stream for convenience. In these cases we return the raw http response to you. The output can be retrieved via `response.content`, `response.iter_lines()` or `response.iter_content()`.
```
amikofalvy:
```suggestion
Any output format can be retrieved as a stream. In these cases we return the raw http response to you. The output can be retrieved via `response.content`, `response.iter_lines()` or `response.iter_content()`.
```

src/rev_ai/apiclient.py
```diff
@@ -195,19 +193,30 @@ def get_transcript_text(self, id_, filename=None, filepath=None):
...
+    def get_transcript_text_as_stream(self, id_):
+        """Get the transcript of a specific job as a plain text stream.
+
+        :param id_: id of job to be requested
+        :returns: requests.models.Response HTTP response which can be used to stream
+            the payload of the response
+        :raises: HTTPError
+        """
+        if not id_:
+            raise ValueError('id_ must be provided')
```
amikofalvy: do we need to note this `:raises: ValueError`? What is the convention?

src/rev_ai/apiclient.py
```diff
@@ -195,19 +193,30 @@ def get_transcript_text(self, id_, filename=None, filepath=None):
...
+
+        url = urljoin(self.base_url, 'jobs/{}/transcript'.format(id_))
```
amikofalvy:
```suggestion
        url = urljoin(self.base_url, 'jobs/{id}/transcript'.format(id_))
```

Are we supposed to put a key in the format bracket?

src/rev_ai/apiclient.py
```diff
@@ -260,13 +280,26 @@ def get_captions(self, id_, filename=None, filepath=None):
...
+        :param id_: id of job to be requested
+        :returns: requests.models.Response HTTP response which can be used to stream
+            the payload of the response
+        :raises: HTTPError
+        """
+        if not id_:
+            raise ValueError('id_ must be provided')
+
+        url = urljoin(self.base_url, 'jobs/{}/captions'.format(id_))
+        response = self.session.get(
```
amikofalvy: Should we do like we have done in the node sdk and separate out an api helper?

amikofalvy: It seems like our python sdk (especially the tests suite) would benefit from the same design pattern.

src/rev_ai/apiclient.py
```diff
@@ -260,13 +280,26 @@ def get_captions(self, id_, filename=None, filepath=None):
...
 
-        if filename:
-            path = os.path.join(filepath or '', filename+'.txt')
-            with open(path, 'w+') as f:
-                f.write(response.text)
-
         return response.text
 
+    def get_captions_as_stream(self, id_):
+        """Get the captions output of a specific job and return it as a plain text stream
```
amikofalvy: Do we have an issue to add vtt output format to python SDK?

src/rev_ai/apiclient.py
k-weng: I believe it is not required by Python.

src/rev_ai/apiclient.py
kbridbur: It's not strictly necessary. My opinion is that it's a good convention to have just to make it clear what arguments are needed and have better error handling if a customer messes up and passes us something undefined

src/rev_ai/apiclient.py
kbridbur: I don't think so. All the other .formats work like this

src/rev_ai/apiclient.py
kbridbur: I will file one. It will likely be done along with AISDK-52 which will abstract away the requests package

PR #29, opened 2019-07-01
AISDK-48-v2: Fix async examples for python SDK

Ymardini:


JDongian (commented):
Is it possible or sensible to have automated tests for these? They would serve as good smoke tests.

kbridbur:
QA is working on integration tests. I would think that would be a good place to have an automated test for these

kbridbur (approved):
:shipit:

PR #30, opened 2019-07-11
AISDK-34: Errors from the API should come back as intended

Ymardini:
Haven't written any tests for it yet. I wanted to see if the implementation was good before writing a test. 
image This is an example of what is returned now when attempting to get the account of an incorrect access token.


src/rev_ai/apiclient.py
```diff
@@ -324,6 +325,18 @@ def get_account(self):
         """
         url = urljoin(self.base_url, 'account')
         response = self.session.get(url)
-        response.raise_for_status()
+        self._api_failure_handler(response)
 
         return Account.from_json(response.json())
+
+    def _api_failure_handler(self, response):
```
kbridbur: This is named as if it's a class. Functions should be named as actions
ie `_handle_api_failure`

tests/test_failure_handler.py
```diff
@@ -0,0 +1,32 @@
...
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
+
+JOB_ID = '1'
+URL = urljoin(RevAiAPIClient.base_url, 'jobs/{}/transcript'.format(JOB_ID))
+
+
+@pytest.mark.usefixtures('mock_client', 'make_mock_response')
+class TestAPIFailureHandler():
```
kbridbur: Just for clarity please change this to be `TestAPIFailureHandling`

tests/test_failure_handler.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
kbridbur: I would normally say to remove this class since you're technically testing a private method which should never be called directly. But I think in this case it's fine. Eventually we should separate direct interaction with the api and failure handling into a separate class that this one can use

src/rev_ai/apiclient.py
```diff
@@ -3,6 +3,7 @@
 
 import requests
 import json
+import sys
```
kbridbur: import no longer needed

tests/test_failure_handler.py
kbridbur: On second thought, I do think we should remove this test class for now. We're currently double testing this logic between the two classes and this class isn't as close to what we will eventually be doing as I thought

tests/test_job.py
```diff
@@ -51,7 +51,7 @@ def test_get_job_details_with_no_job_id(self, id, mock_client):
         ['unauthorized', 'job-not-found']))
     def test_get_job_details_with_error_response(self, error, mock_client, make_mock_response):
         status = error.get('status')
-        response = make_mock_response(url=JOB_ID_URL, status=status, json_data=error)
+        response = make_mock_response(url=JOB_ID_URL, status=status, json_data=error, text=status)
```
kbridbur: To make sure we handle the message text properly I would set text to something else along with the match parameter on `pytest.raises`

kbridbur: Please do the same for all the similar tests

kbridbur (changes requested):
Some comments

src/rev_ai/apiclient.py
JDongian: flake8 can detect this, make sure to run that before shipping

src/rev_ai/apiclient.py
```diff
@@ -324,6 +325,18 @@ def get_account(self):
...
+    def _handle_api_failure(self, response):
+        """Helper function to provide more detailed error messages for the user
+        
+        :param response: response from the HTTP request
+        :raises: HTTPError
+        """
+        try:
+            response.raise_for_status()
+        except Exception as err:
+           err.args = (response.text,)
```
JDongian: Why are we returning the JSON?
```
HTTPError: {"title":"Authorization has been denied for this request.","status":401}
```
doesn't seem better than then vanilla
```
HTTPError: 401 Client Error: Unauthorized for url: https://api.rev.ai/speechtotext/v1/account
```
to me.

src/rev_ai/apiclient.py
```diff
@@ -324,6 +325,18 @@ def get_account(self):
         """
         url = urljoin(self.base_url, 'account')
         response = self.session.get(url)
-        response.raise_for_status()
+        self._handle_api_failure(response)
```
JDongian: An alternate pattern you could have used:
```
@rev_http_wrapper
def get_account(self):
    url = urljoin(self.base_url, 'account')
    response = self.session.get(url)
    response.raise_for_status()

    return Account.from_json(response.json())
```

where you have rev_http_wrapper do your thing with the exception handling.

Why would I choose this over the inline changes? The decorator provides some abstraction -- this keeps the method body simpler with familiar raise_for_status() calls and tells readers not to worry about the http handling behavior. After all, we are more concerned with capturing the HTTPError than the actual behavior of `raise_for_status`. Also, if you want to modify the http error handling behavior in the future, you have a bit more power with a decorator than with the `self._handle_api_failure(response)` method call. You should weigh this against your reasons to use a private method to wrap the `raise_for_status` call.

src/rev_ai/apiclient.py
```diff
@@ -324,6 +325,18 @@ def get_account(self):
...
+
+    def _handle_api_failure(self, response):
+        """Helper function to provide more detailed error messages for the user
+        
+        :param response: response from the HTTP request
+        :raises: HTTPError
+        """
+        try:
+            response.raise_for_status()
+        except Exception as err:
```
JDongian: Prefer to use as specific exceptions as possible. Reasons: code-as-documentation and better clarity in logs in case of other kind of exception.

src/rev_ai/apiclient.py
kbridbur: In this case it's not better than vanilla but if we were to take 409 as an example (we return this code when a customer tries to get the transcript of an in progress job) it would be 
`HTTPError: 409 Client Error: Conflict: https://api.rev.ai/speechtotext/v1/jobs/.../transcript`
which is much worse than 
`HTTPError: {
    "allowed_values": [
        "transcribed"
    ],
    "current_value": "in_progress",
    "type": "https://www.rev.ai/api/v1/errors/invalid-job-state",
    "title": "Job is in invalid state",
    "status": 409,
    "detail": "Job is in progress or failed"
}`

src/rev_ai/apiclient.py
kbridbur: This decorator abstraction is definitely nice, I'll have to look more into decorators in python. I have future plans to completely abstract the response library and error handling from this client in the future (similar to what we did for the node sdk if you've seen that). 

Maybe it makes sense to just do that in this issue

src/rev_ai/apiclient.py
kbridbur: There definitely shouldn't be any difference in the amount of information the customer has access to between the api and the sdks. In this case the two options that come to mind are to do this or to return the json object, which I definitely don't think we should do as it breaks the promise of the method

examples/microphone_streaming_example.py
```diff
@@ -83,6 +81,7 @@ def generator(self):
 
             yield b''.join(data)
 
+
```
kbridbur, line 84: Did the linter tell you to add this line?

src/rev_ai/apiclient.py
```diff
@@ -322,8 +341,29 @@ def get_account(self):
...
+            and stream
+        :raises: HTTPError
+        """
+        response = self.session.request(method, url, **kwargs)
+
+        try:
+            response.raise_for_status()
+            return response
+        except HTTPError as err:
+            err.args = (err.args[0] +
```
kbridbur: We should make sure response.content exists. Currently if they get a 500 or a 503 it may error

tests/test_apiclient.py
```diff
@@ -21,3 +23,16 @@ def test_constructor_with_success(self):
...
+    @pytest.mark.parametrize('error', get_error_test_cases(
+        ['unauthorized', 'job-not-found', 'invalid-job-state']))
+    @pytest.mark.parametrize('method', ["POST", "GET", "DELETE"])
+    def test_HTTP_handler(self, error, method, mock_client, make_mock_response):
+        status = error.get('status')
+        URL = RevAiAPIClient.base_url
+        response = make_mock_response(url=URL, status=status, json_data=error)
+        mock_client.session.request.return_value = response
+
+        with pytest.raises(HTTPError, match=str(status)):
```
kbridbur: We should improve this match to make sure the error is also included

src/rev_ai/apiclient.py
```diff
@@ -322,8 +341,29 @@ def get_account(self):
...
-        response = self.session.get(url)
-        response.raise_for_status()
+        response = self._HTTPHandler(
+                      "GET",
+                      urljoin(self.base_url, 'account')
+        )
 
         return Account.from_json(response.json())
+
+    def _HTTPHandler(self, method, url, **kwargs):
```
kbridbur: This is still named as if it's a class (a noun). It should be named like a method (a verb). Something like `_make_http_request`

src/rev_ai/apiclient.py
```diff
@@ -322,8 +341,29 @@ def get_account(self):
...
+        """Wrapper method for initiating HTTP requests and handling potential
```
kbridbur, line 352: I like that you documented this, I'm all for more documentation. But just for future reference, since it's a private method it doesn't really NEED to be documented

tests/test_captions.py
```diff
@@ -2,7 +2,6 @@
 
 import pytest
 import os.path
-from requests.exceptions import HTTPError
 from src.rev_ai.apiclient import RevAiAPIClient
 from tests.helpers.errors import get_error_test_cases
```
kbridbur: I don't think this import is needed anymore

tests/test_job.py
```diff
@@ -3,7 +3,6 @@
 
 import json
 import pytest
-from requests.exceptions import HTTPError
 from src.rev_ai.apiclient import RevAiAPIClient
 from src.rev_ai.models.asynchronous import Job, JobStatus
 from tests.helpers.errors import get_error_test_cases
```
kbridbur: same thing

kbridbur (approved):
Looking good. Make sure you run the linter and install it locally to test. No major comments.

tests/test_apiclient.py
```diff
@@ -21,3 +23,16 @@ def test_constructor_with_success(self):
     def test_constructor_with_no_token(self, token):
         with pytest.raises(ValueError, match='access_token must be provided'):
             RevAiAPIClient(token)
+
+    @pytest.mark.parametrize('error', get_error_test_cases(
+        ['unauthorized', 'job-not-found', 'invalid-job-state']))
+    @pytest.mark.parametrize('method', ["POST", "GET", "DELETE"])
```
kbridbur, line 31: Do these test cases run as a matrix? ie. each error is run with each method? If so that's pretty convenient :)

examples/microphone_streaming_example.py
Ymardini: Yep.

tests/test_apiclient.py
Ymardini: Yes they do! :)

kbridbur (approved):
Once the tests pass :shipit:

PR #32, opened 2019-07-30
Update streamingclient.py

JDongian:
https://tools.ietf.org/html/rfc1738 section 3.3
>  Within the <path> and <searchpart> components, "/", ";", "?" are reserved.

So you should be encoding it even though the server is lenient.

Also, this is the first time I've seen a content type in a query string. Is that a websocket thing?

### DIFF
--- a/src/rev_ai/streamingclient.py
+++ b/src/rev_ai/streamingclient.py
@@ -4,6 +4,10 @@ import websocket
 import threading
 import six
 import json
+try:
+    from urllib import urlencode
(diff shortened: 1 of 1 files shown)


kbridbur:
Looks like a good idea, please update the one test which is failing

To answer your question. Content type is in the query string because many javascript (and maybe some other languages) websocket libraries don't support headers.

kbridbur:
Addressed in another PR

PR #33, opened 2019-08-12
AISDK-55: Catch SDK up to API in features

kbridbur:


README.md
```diff
@@ -115,10 +115,11 @@ just the text of your transcript
 
 ### Getting captions output
 
-You can also get captions output from the SDK:
+You can also get captions output from the SDK. We offer both SRT and VTT caption formats.
+If you submitted your job as a multichannel then you must also provide a `channel_id` to be captioned:
```
amikofalvy: I think we refer to it as speaker channels.

Speaker Channels = 1 speaker per channel
Multi Channel = generic term for audio with multiple channels (example: stereo audio)

```suggestion
If you submitted your job as a speaker channel then you must also provide a `channel_id` to be captioned:
```

src/rev_ai/apiclient.py
```diff
@@ -64,6 +63,10 @@ def submit_job_url(
         :param metadata: info to associate with the transcription job
         :param callback_url: callback url to invoke on job completion as a webhook
         :param skip_diarization: should rev.ai skip diaization when transcribing this file
+        :param skip_punctuation: should rev.ai skip punctuation when transcribing this file
+        :param speaker_channel_count: the number of speaker channels in the audio. If provided
+            the given audio will have each channel transcribed separately and each channel
+            will be treated as a single speaker. [1-8]
```
amikofalvy:
```suggestion
            will be treated as a single speaker. Valid values are integers 1 through 8 inclusive.
```

src/rev_ai/apiclient.py
```diff
@@ -106,6 +115,10 @@ def submit_job_local_file(
...
```
amikofalvy: I suggest the same changes as above.

src/rev_ai/apiclient.py
```diff
@@ -120,17 +133,21 @@ def submit_job_local_file(
...
         if custom_vocabularies:
             payload['custom_vocabularies'] = custom_vocabularies
+        if speaker_channel_count:
+            payload['speaker_channel_count'] = speaker_channel_count
 
         with open(filename, 'rb') as f:
             files = {
                 'media': (filename, f),
-                'options': (None, json.dumps(payload))
+                'options': (None, json.dumps(payload, sort_keys=True))
```
amikofalvy, line 130: Why sort_keys?

src/rev_ai/models/asynchronous/caption_type.py
```diff
@@ -0,0 +1,13 @@
...
+
+from enum import Enum
+
+
+class CaptionType(Enum):
+    SRT = "application/x-subrip"
+    VTT = "text/vtt"
+
+    @classmethod
+    def from_string(cls, status):
```
amikofalvy, line 12: is this an override of a default python method?

Makefile
```diff
@@ -51,7 +51,8 @@ clean-test: ## remove test and coverage artifacts
 	rm -fr .pytest_cache
 
 lint: ## check style with flake8
-	flake8 rev_ai tests
+	flake8 rev_ai tests --ignore=F401,W504,E731
+	flake8 rev_ai src --ignore=F401,W504
```
amikofalvy, line 55: Why no E731?

src/rev_ai/apiclient.py
kbridbur: I was having trouble getting the tests to pass because our testing sorts in some cases and doesn't in others. This was the best way I could figure out to make it consistent without much change

src/rev_ai/models/asynchronous/caption_type.py
kbridbur: No it is not. This was automatically added when I built actually. You can see the same in the job_status enum

src/rev_ai/apiclient.py
```diff
@@ -120,23 +133,27 @@ def submit_job_local_file(
         payload = {}
         if skip_diarization:
             payload['skip_diarization'] = skip_diarization
+        if skip_punctuation:
+            payload['skip_punctuation'] = skip_punctuation
```
amikofalvy: Can this logic be extracted into a payload_creator and used by both submit job methods?

Makefile
kbridbur: https://lintlyci.github.io/Flake8Rules/rules/E731.html

This makes some sense in a public api where everything needs to be highly readable. Assigning lambdas is a nice feature to have for testing though

src/rev_ai/apiclient.py
```diff
@@ -272,47 +289,57 @@ def get_transcript_object(self, id_):
...
+        :param channel_id: id of speaker channel to be captioned, only matters for multichannel jobs
         :returns: requests.models.Response HTTP response which can be used to stream
             the payload of the response
         :raises: HTTPError
         """
         if not id_:
             raise ValueError('id_ must be provided')
+        query = ""
+        if channel_id:
+            query = "?speaker_channel={}".format(channel_id)
```
amikofalvy: same question about extracting shared logic into helpers

kbridbur: Moving this to a shared helper seems unnecessary. It's not much duplicate code and the helper function is not useful on its own.

.travis.yml
```diff
@@ -11,4 +11,6 @@ matrix:
       dist: xenial
       sudo: true
 install: pip install -r requirements.txt -r requirements_dev.txt
-script: make test
+script: 
+  - make test
+  - make lint
```
kbridbur, line 16: Added linting as a step in travis

Makefile
kbridbur, line 55: Rules can be searched here: https://lintlyci.github.io/Flake8Rules/

tests/test_captions.py
```diff
@@ -14,27 +15,59 @@
...
+        res = mock_client.get_captions(JOB_ID, content_type)
         assert res == data
         mock_client.session.request.assert_called_once_with(
             "GET",
             URL,
-            headers={'Accept': 'application/x-subrip'}
+            headers={'Accept': expected_content_type}
+        )
+
+    def test_get_captions_with_speaker_channel(self, mock_client, make_mock_response):
```
amikofalvy, line 47: Does this test assert anything different from `test_get_captions_default_content_type`?

src/rev_ai/apiclient.py
amikofalvy: I worry about future developers not implementing changes in both locations.

tests/test_captions.py
```diff
@@ -14,27 +15,59 @@
...
+        expected_content_type = CaptionType.SRT.value
         response = make_mock_response(url=URL, text=data)
         mock_client.session.request.return_value = response
 
         res = mock_client.get_captions(JOB_ID)
+        assert res == data
+        mock_client.session.request.assert_called_once_with(
+            "GET",
+            URL,
+            headers={'Accept': expected_content_type}
```
amikofalvy, line 29: Do you remember how we refactored the node sdk to put the HTTP call in another layer so that we could make easier assertions? Should we do something similar here?

It appears that every small change to one of our method that touches an endpoint forces us to make a large number of line edits in the test file.

tests/test_job.py
```diff
@@ -96,12 +96,16 @@ def test_submit_job_url_with_success(self, mock_client, make_mock_response):
             'status': 'in_progress',
             'created_on': CREATED_ON,
             'metadata': METADATA,
-            'callback_url': CALLBACK_URL
+            'callback_url': CALLBACK_URL,
```
amikofalvy, line 99: the other test has `skip_diarization` but this one does not. Should we add it?

tests/test_captions.py
kbridbur: We already pretty much do this in self._make_http_request. It's essentually the same as we have in the python SDK. I don't think it would cut down on tests realistically

tests/test_captions.py
kbridbur: It asserts that the url is correct when we add the speaker channel specification

README.md
```diff
@@ -115,10 +115,11 @@ just the text of your transcript
 
 ### Getting captions output
 
-You can also get captions output from the SDK:
+You can also get captions output from the SDK. We offer both SRT and VTT caption formats.
+If you submitted your job as speaker channeled audio then you must also provide a `channel_id` to be captioned:
```
amikofalvy:
```suggestion
If you submitted your job as speaker channel audio then you must also provide a `channel_id` to be captioned:
```

amikofalvy: Is this a word?

tests/test_captions.py
kbridbur: I think the real cause is that we have `_as_stream` and normal versions of each output method. Essentially doubling the number of tests

README.md
kbridbur: I'm not sure but I think it gets the point across

amikofalvy (approved):
:shipit:

PR #34, opened 2019-08-22
 AISDK-59: Create new session for each request

kbridbur:


src/rev_ai/apiclient.py
```diff
@@ -42,11 +42,10 @@ def __init__(self, access_token):
         if not access_token:
             raise ValueError('access_token must be provided')
 
-        self.session = requests.Session()
-        self.session.headers.update({
+        self.default_headers = {
```
kbridbur, line 45: The changes to this file are the important ones. We no longer have a persistent session for all requests

src/rev_ai/apiclient.py
```diff
@@ -361,7 +360,12 @@ def _make_http_request(self, method, url, **kwargs):
             and stream
         :raises: HTTPError
         """
-        response = self.session.request(method, url, **kwargs)
+        headers = self.default_headers.copy()
+        if ('headers' in kwargs):
```
JDongian:
```suggestion
        if 'headers' in kwargs:
```

src/rev_ai/apiclient.py
```diff
@@ -361,7 +360,12 @@ def _make_http_request(self, method, url, **kwargs):
...
+            headers.update(kwargs.get('headers'))
+            del kwargs['headers']
```
JDongian, line 366: I think this solution is fine. We don't expect to actually modify the headers often, right?

tests/test_apiclient.py
```diff
@@ -29,15 +29,18 @@ def test_constructor_with_no_token(self, token):
...
         ['unauthorized', 'job-not-found', 'invalid-job-state']))
     @pytest.mark.parametrize('method', ["POST", "GET", "DELETE"])
-    def test_make_http_request(self, error, method, mock_client, make_mock_response):
+    def test_make_http_request(self, error, method, mock_session, make_mock_response):
         status = error.get('status')
         URL = RevAiAPIClient.base_url
         response = make_mock_response(url=URL, status=status, json_data=error)
-        mock_client.session.request.return_value = response
+        mock_session.request.return_value = response
+        client = RevAiAPIClient("token")
```
JDongian: I think there's some inconsistency that could be cleaned up with the usage of TOKEN, token, and "token"

tests/test_captions.py
```diff
@@ -10,112 +10,134 @@
...
 
-    def test_get_captions_with_speaker_channel(self, mock_client, make_mock_response):
+    def test_get_captions_with_speaker_channel(self, mock_session, make_mock_response):
         data = 'Test'
         channel_id = 1
         expected_url = URL + '?speaker_channel={}'.format(channel_id)
         expected_content_type = CaptionType.SRT.value
+        client = RevAiAPIClient(TOKEN)
+        expected_headers = {'Accept': expected_content_type}
+        expected_headers.update(client.default_headers)
```
JDongian, line 63: Probably not a problem, but I would still be careful that we are conscious that the `_make_http_request` implementation overrides the defaults with the kwarg, but here in the tests we generate the expectations the reverse way.

JDongian (approved):
looks reasonable

tests/test_apiclient.py
kbridbur: I missed this, I should be using TOKEN everywhere

src/rev_ai/apiclient.py
kbridbur: The only time's we have to do this is when we're trying to get output, the content type header needs to be added

tests/test_captions.py
kbridbur: It's slightly unfortunate. I think in the future we should separate `_make_http_request` into a new class and refactor the tests. Then we won't have to do any modification of arguments in tests

PR #40, opened 2019-10-28
AISDK-69: Add Client for interacting with the Custom Vocab API Python

KostasRev:
Current status: both getting and submitting and getting custom vocabularies works. Need to add unit tests.


src/rev_ai/customvocabulariesclient.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class RevAiCustomVocabulariesClient:
+    """Client which implements Rev.ai CustomVocabulary API"""
+
+    # Default version of Rev.ai
+    version = 'v1'
+
+    # Default address of the API
+    base_url = 'https://api.rev.ai/speechtotext/{}/vocabularies'.format(
```
amikofalvy: @kbridbur , Do we want to drop Python 2.7 support? If we do, we can start to use python fstrings

src/rev_ai/customvocabulariesclient.py
```diff
@@ -0,0 +1,125 @@
...
+                             settings page of your account dashboard
+                             on Rev.ai
+        """
+        if not access_token:
+            raise ValueError('access_token must be provided')
+
+        self.default_headers = {
+            'Authorization': 'Bearer {}'.format(access_token),
+            'User-Agent': 'RevAi-PythonSDK/{}'.format(__version__)
+        }
```
amikofalvy: These are so similar to the apiclient.py class definition. Why do you choose to define the below functions in a new file instead of apiclient.py?

src/rev_ai/customvocabulariesclient.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def submit_custom_vocabularies(
+            self,
+            custom_vocabularies,
```
amikofalvy: Another question: Should we define a Python object for custom_vocabularies or do we leave it open as generic object?

In my opinion, SDKs are developer guides and guides don't leave anything open for interpretation.

amikofalvy: But I hear from Kyle that it is Python standard to use raw objects instead of named object wrappers. You should talk to Kyle to make the final decision on this.

src/rev_ai/customvocabulariesclient.py
kbridbur: We should continue with Python2.7 support until the end of its lifetime (Jan 1st 2020)

src/rev_ai/customvocabulariesclient.py
kbridbur: I think it's a good idea to have it as a separate client (especially since these are not usable in the normal api currently, having them in the same client will lead to confusion)

I would do something similar to what we did in the node SDK and create a base class which both apiclient and customvocab client can extend. We do something similar in the node SDK

src/rev_ai/customvocabulariesclient.py
kbridbur: I think in this case we should make this a list of "CustomVocabObjects". I guess to clarify a bit more, as far as I can tell, in python it's not normal to wrap `options` or compound objects in a single object to be used as an argument. Arguments are usually built in types or close to built in types (int, string, list). 

In this case `custom_vocabularies` is essentially a list of dictionaries, since all the dictionaries have the same key: `phrases` I think it makes sense to wrap that dictionary in an object

I don't know the formal reasoning for this but I would guess it's because python lets you use named arguments in functions so having a wrapper object around options doesn't really simplify anything like it does in node

src/rev_ai/customvocabulariesclient.py
```diff
@@ -0,0 +1,125 @@
...
+        payload = {}
+        if callback_url:
+            payload['callback_url'] = callback_url
+        if metadata:
+            payload['metadata'] = metadata
+        if custom_vocabularies:
+            payload['custom_vocabularies'] = custom_vocabularies
+        return payload
+
+    def _make_http_request(self, method, url, **kwargs):
```
kbridbur: As I mentioned above. I think it would be a good idea to either make a base class or a class which both this and the `apiclient` can depend on which handles the actual creation of api requests (creating the auth header, disposing the http session). We do something similar in the node sdk here: https://github.com/revdotcom/revai-node-sdk/blob/develop/src/api-request-handler.ts

Benefits of this are: reduced duplicate code and easier testing

examples/custom_vocabularies_example.py
```diff
@@ -0,0 +1,36 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from rev_ai import customvocabulariesclient
+
+access_token = "your_access_token"
+
+print("Creating Custom Vocab Client")
```
kbridbur, line 22: I think you can reduce the number of new lines in this file. Newlines only really need to go between "logical blocks" meaning sections of code with different purposes

examples/custom_vocabularies_example.py
```diff
@@ -0,0 +1,36 @@
...
+
+print("Submitting")
+
+custom_vocab = client.submit_custom_vocabularies([{"phrases":["Peace out"]}])
+
+print("Succeeded submission")
+
+print("Getting Custom Vocabs")
+
+result = client.get_custom_vocabularies(custom_vocab["id"])
```
kbridbur: I think it would be nice if we did something similar to the async example here and had a waiting loop for this custom vocab to be ready to use

kbridbur (commented):
You'll also need to add to the README

examples/custom_vocabularies_example.py
```diff
@@ -0,0 +1,52 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import time
+from rev_ai import customvocabulariesclient
```
amikofalvy: @kbridbur What is python convention for naming clients? I'm afraid this client name is hard to read due to a lack of formatting.

snake case?

examples/custom_vocabularies_example.py
```diff
@@ -0,0 +1,52 @@
...
+
+access_token = "your_access_token"
+
+print("Creating Custom Vocab Client")
+client = customvocabulariesclient.RevAiCustomVocabulariesClient(access_token)
+
+print("Submitting")
+custom_vocabularies_job = client.submit_custom_vocabularies(
+    [
+        {"phrases": ["Peace out"]},
```
amikofalvy: I think we should use more professional terms in our example. Let's use the names of well known members of the tech community that have unique names.

amikofalvy: In our documentation for the Async API we use this example:
```
"custom_vocabularies": [
{
"phrases": [
"Paul McCartney",
"Amelia Earhart"
]
}
],
```

examples/custom_vocabularies_example.py
```diff
@@ -0,0 +1,52 @@
...
+        CustomVocabulary(["What Up", "skeddadle"]).get_raw()
```
amikofalvy: I wouldn't expect the customer to have to call get_raw().

In our zoom conversation yesterday, I thought we decided that `submit_custom_vocabularies` would handle each type independently. If the type is a CustomVocabulary object the `submit_custom_vocabularies` should call `get_raw` internally.

Also, I think `get_raw`  should instead be called `to_dict()`

src/rev_ai/__init__.py
```diff
@@ -3,4 +3,5 @@
 
 __version__ = '2.6.1'
 
-from .models import Job, JobStatus, Account, Transcript, MediaConfig, CaptionType
+from .baseclient import BaseClient
```
amikofalvy: BaseClient should be internal only. Does this syntax make it so that customers can access BaseClient?

src/rev_ai/baseclient.py
```diff
@@ -0,0 +1,72 @@
...
+            if (response.content):
+                err.args = (err.args[0] +
+                            "; Server Response : {}".
+                            format(response.content.decode('utf-8')),)
+            raise
+
+    def _process_vocabularies(self, unprocessed_vocabularies):
+        processed_vocabularies = []
+        for custom_vocabulary in unprocessed_vocabularies:
+            processed_vocabularies.append(custom_vocabulary.get_raw() if isinstance(custom_vocabulary, CustomVocabulary) else custom_vocabulary)
```
amikofalvy: why no mapping function?

src/rev_ai/__init__.py
KostasRev: I only made this change in order to make imports be from `.` rather than from `.baseclient` in `apiclient.py` and `customvocabulariesclient.py`. But I see your point in keeping it internal.

In the same vein, would we want to put the other three api clients in this `__init__ `for easier access?

examples/custom_vocabularies_example.py
amikofalvy: I guess this isn't a client name, but rather a module name

examples/custom_vocabularies_example.py
```diff
@@ -0,0 +1,51 @@
...
+
+access_token = "your_access_token"
+
+print("Creating Custom Vocab Client")
+client = customvocabulariesclient.RevAiCustomVocabulariesClient(access_token)
+
+print("Submitting")
+custom_vocabularies_job = client.submit_custom_vocabularies(
+    [
+        {"phrases": ["Patrick Henry Winston"]},
```
amikofalvy: I think that these names are common enough that our engine would correctly identify them without custom vocabulary.

May I suggest more difficult names like: `Steve Wozniak`

kbridbur: class names are PascalCase

src/rev_ai/baseclient.py
```diff
@@ -0,0 +1,69 @@
...
+            return response
+        except HTTPError as err:
+            if (response.content):
+                err.args = (err.args[0] +
+                            "; Server Response : {}".
+                            format(response.content.decode('utf-8')),)
+            raise
+
+    def _process_vocabularies(self, unprocessed_vocabularies):
+        return list(map(lambda custom_vocabulary: custom_vocabulary.to_dict() if isinstance(custom_vocabulary, CustomVocabulary) else custom_vocabulary, unprocessed_vocabularies))
```
amikofalvy: minor - long line please wrap

src/rev_ai/customvocabulariesclient.py
amikofalvy: I think Kostas has arrive at a solution, @kbridbur if this is satisfactory will you resolve this comment?

src/rev_ai/customvocabulariesclient.py
```diff
@@ -0,0 +1,85 @@
...
+            "POST",
+            self.base_url,
+            json=payload
+        )
+
+        return response.json()
+
+    def get_custom_vocabularies(
+            self,
+            id):
```
amikofalvy: @KostasRev are you running the python linter?

src/rev_ai/customvocabulariesclient.py
```diff
@@ -0,0 +1,85 @@
...
+        :param access_token: access token which authorizes all requests and
+                             links them to your account. Generated on the
+                             settings page of your account dashboard
+                             on Rev.ai
+        """
+        BaseClient.__init__(self, access_token)
+
+        self.base_url = urljoin(self.base_url, 'vocabularies/')
+
+    def submit_custom_vocabularies(
```
amikofalvy: I think we are missing unit tests for this client

examples/custom_vocabularies_example.py
kbridbur: module names should be all lowercase, however it would be useful to insert some `_`s to make this more readable

kbridbur: so snake_case I guess

src/rev_ai/customvocabulariesclient.py
kbridbur: The linter can be run by running `python setup.py lint` from the main directory or manually by running `flake8`

amikofalvy:
Is this status still accurate?
https://github.com/revdotcom/revai-python-sdk/pull/40#issue-333355380

README.md
```diff
@@ -172,4 +172,28 @@ If you want to end the connection early, you can!
...
+custom_vocabulary = CustomVocabulary(["Patrick Henry Winston", "Robert Berwick", "Noam Chomsky"])
+
+# Submit the CustomVocabulary
+custom_vocabularies_job = client.submit_custom_vocabularies([custom_vocabulary])
+
+# View the job's progress
+job_state = client.get_custom_vocabularies(custom_vocabularies_job['id'])
+```
+
+For more details, check out the custom vocabularies example.
```
kbridbur: Lets link to the directory with the example here

examples/async_example.py
```diff
@@ -59,8 +61,8 @@
 
     if status == "TRANSCRIBED":
         # Getting a list of current jobs connected with your account
-        # The optional parameters limits limits the length of the list. Starting_after
-        # Cuts off the beginning x jobs off the list returned
+        # The optional parameters limits the length of the list.
+        # Starting_after cuts off the beginning x jobs off the list returned
```
kbridbur: Not your fault but this xml doc is wrong. should be something like
`starting_after is a job id which causes the removal of all jobs from the list which were created before that job`

src/rev_ai/baseclient.py
```diff
@@ -0,0 +1,65 @@
...
+            response.raise_for_status()
+            return response
+        except HTTPError as err:
+            if (response.content):
+                err.args = (err.args[0] +
+                            "; Server Response : {}".
+                            format(response.content.decode('utf-8')),)
+            raise
+
+    def _process_vocabularies(self, unprocessed_vocabularies):
```
kbridbur: This is specific to the custom vocab client and should probably go there

kbridbur: Please write a xml doc for this. It's not immediately clear what it's doing

src/rev_ai/apiclient.py
```diff
@@ -60,16 +48,21 @@ def submit_job_url(
...
-        :param skip_diarization: should rev.ai skip diaization when transcribing this file
-        :param skip_punctuation: should rev.ai skip punctuation when transcribing this file
-        :param speaker_channels_count: the number of speaker channels in the audio. If provided
-            the given audio will have each channel transcribed separately and each channel
-            will be treated as a single speaker. Valid values are integers 1-8 inclusive.
-        :param custom_vocabularies: a collection of phrase dictionaries. Including custom
-            vocabulary will inform and bias the speech recognition to find those phrases.
-            Each dictionary should consist of a key "phrases" which maps to a list of strings,
-            each of which represents a phrase you would like the speech recognition to bias
+        :param callback_url: callback url to invoke on job completion as
```
kbridbur, line 52: was flake complaining about these lines?

src/rev_ai/baseclient.py
kbridbur: This method also doesn't feel at home in this class to me. If this is a BaseClient it could presumably be extended for something that has nothing to do with custom vocabs. I would rather this be moved to a helper class. Now might be a useful time to create a utils.py or helpers.py file and put this in it

src/rev_ai/custom_vocabularies_client.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+
+        response = self._make_http_request(
+            "GET",
+            urljoin(self.base_url, "{}".format(id))
+        )
+
+        return response.json()
+
+    def _create_custom_vocabularies_options_payload(
```
kbridbur, line 72: I think it's somewhat unnecessary to separate this into another method. This method isn't going to be used anywhere else and it doesn't make the caller less complex

It's somewhat preference though. How it is isn't bad

src/rev_ai/apiclient.py
```diff
@@ -1,34 +1,27 @@
...
+from .baseclient import BaseClient
 
 try:
     from urllib.parse import urljoin
 except ImportError:
     from urlparse import urljoin
 
 
-class RevAiAPIClient:
+class RevAiAPIClient(BaseClient):
```
kbridbur, line 15: I wonder if we can simplify the tests now that we have this separation. In theory we should be able to mock out the `_make_http_request` call and reduce a lot of duplicated checks in the tests 

@k-weng how easy do you think it would be to do this

KostasRev: I'm actually currently modifying the tests for `_make_http_request`. Removed from `test_apiclient` and moved it to `test_baseclient`. It makes sense to me and seems to work just fine. Is that what you mean @kbridbur ?

src/rev_ai/custom_vocabularies_client.py
KostasRev: It makes it better for understanding of the http request structure as well as being more modular which makes the code easier to understand in my opinion. It also parallels the pattern in `apiclient` which seems nice, but is no solid reason on its own.

src/rev_ai/apiclient.py
KostasRev: I suppose not, but Sublime's PEP8 plugin I'm running complained, which is why I changed them. As far as I can tell, PEP8 applies to the docstrings' length as well.

src/rev_ai/baseclient.py
KostasRev: This makes sense given that the streaming client also could use this method in the future.

src/rev_ai/__init__.py
KostasRev: Resolving as I believe it has been addressed and you have seen the resolution and agree with it.

README.md
```diff
@@ -172,4 +172,28 @@ If you want to end the connection early, you can!
...
+In this example you see how to construct custom vocabulary objects, submit them to the api, and check on their progress!
+
+```python
+from rev_ai import custom_vocabularies_client
+from rev_ai.models import CustomVocabulary
+# Create a client
+client = custom_vocabularies_client.RevAiCustomVocabulariesClient("ACCESS TOKEN")
+
+# Construct a CustomVocabulary object using your desired phrases
+custom_vocabulary = CustomVocabulary(["Patrick Henry Winston", "Robert C Berwick", "Noam Chomsky"])
```
JohnDonavon, line 192: The input to the custom vocabulary input should be a list of objects each containing a list of phrases.
i.e.:
```
"custom_vocabularies": [
		{
			"phrases": ["Sparkletini", "Science stuff"]
		},
		{
			"phrases": ["Barbara Streisand"]
		}
	]
```
I think we should rewrite this example to reflect that.

JohnDonavon: Wait jumped the gun, disregard

src/rev_ai/utils.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+"""Speech recognition tools for using Rev.ai"""
+
+from . import CustomVocabulary
+
+
+def _process_vocabularies(unprocessed_vocabularies):
+    return list(map(lambda custom_vocabulary: custom_vocabulary.to_dict()
+                    if isinstance(custom_vocabulary, CustomVocabulary)
+                    else custom_vocabulary, unprocessed_vocabularies))
```
JohnDonavon, line 16: Should we have a check to make sure that the unprocessed vocab has the key "phrases"?

README.md
```diff
@@ -172,4 +172,28 @@ If you want to end the connection early, you can!
 streaming_client.end()
 ```
 
-Otherwise, the connection will end when the server obtains an "EOS" message.
\ No newline at end of file
+Otherwise, the connection will end when the server obtains an "EOS" message.
+
+### Submitting Custom Vocabularies
+
+In addition to passing custom vocabularies as parameters in the async api client, you can create and view your custom vocabularies independently.
```
amikofalvy: This statement is misleading.

1. You cannot view the custom vocabulary in the way that the customer is probably expecting. You can only view the CV processing status and the metadata, not the CV phrases themselves.
2. There is little motivation for a customer to presubmit CV for async jobs. The primary motivation for presubmitting CV is to use CV with streaming. I think your description should make this clear.

src/rev_ai/apiclient.py
amikofalvy: @KostasRev, what @kbridbur is referring to is the test pattern we use in the revai node sdk. One of the benefits of extracting the baseclient is that we can independently test the baseclient and remove any excessive test assertions for asserting calls to the httpclient directly.

let's zoom tomorrow if we need to discuss it more.

src/rev_ai/utils.py
amikofalvy: Validation should be done by the REST API. It's more flexible for development. Validating by API reduces the amount of validation code we need to rewrite in the event of an API change.

@KostasRev please remove sdk (client) side validation.

src/rev_ai/utils.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def _validate_custom_vocabularies(custom_vocabularies_list):
+    for custom_vocabulary in custom_vocabularies_list:
+        if len(custom_vocabulary.keys()) != 1 or 'phrases' not in custom_vocabulary.keys():
+            raise ValueError("Custom vocabularies malformed, must be dict with \
```
amikofalvy: This is the part that should be removed.

src/rev_ai/utils.py
amikofalvy: The API will return an error with the correct error message in the event the request is malformed.

tests/test_baseclient.py
```diff
@@ -0,0 +1,52 @@
...
+
+class TestBaseClient:
+    def test_constructor_with_success(self):
+        client = BaseClient(TOKEN)
+
+        headers = client.default_headers
+
+        assert headers.get(
+            'User-Agent') == 'RevAi-PythonSDK/{}'.format(__version__)
+        assert headers.get('Authorization') == 'Bearer {}'.format(TOKEN)
```
amikofalvy: In my opinion we should be using string literals in the assertions where possible. String literals are less prone to mistakes.

src/rev_ai/utils.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+"""Speech recognition tools for using Rev.ai"""
+
+from . import CustomVocabulary
+
+
+def _process_vocabularies(unprocessed_vocabularies):
```
amikofalvy, line 7: Missing unit tests for this function

src/rev_ai/models/customvocabulary.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self, phrases):
+        """Constructor
+
+        :param phrases: list of strings of custom phrases to be recognized in
+                        submitted audio
+        """
+        self.phrases = [phrase for phrase in phrases]
+
+    def to_dict(self):
```
amikofalvy, line 14: missing unit tests for this function

src/rev_ai/custom_vocabularies_client.py
```diff
@@ -0,0 +1,87 @@
...
+        if not custom_vocabularies:
+            raise ValueError('custom_vocabularies must be provided')
+
+        payload = self._create_custom_vocabularies_options_payload(
+            custom_vocabularies,
+            callback_url,
+            metadata
+        )
+
+        response = self._make_http_request(
```
amikofalvy, line 50: Is it possible to create a mock of the base client and assert that this function was called with the correct parameters?

We are missing test logic for all function calls in both this cv client and the api client

src/rev_ai/utils.py
KostasRev: Alright, will do.

tests/test_custom_vocabularies_client.py
```diff
@@ -0,0 +1,26 @@
...
+        client = RevAiCustomVocabulariesClient(TOKEN)
+
+        headers = client.default_headers
+
+        assert headers.get(
+            'User-Agent') == 'RevAi-PythonSDK/{}'.format(__version__)
+        assert headers.get('Authorization') == 'Bearer {}'.format(TOKEN)
+
+    @pytest.mark.parametrize('token', [None, ''])
+    def test_constructor_with_no_token(self, token):
```
kbridbur, line 25: We should have tests for each request to make sure they call `BaseClient._make_http_request` with the proper arguments

kbridbur: This might help
https://stackoverflow.com/questions/12219967/how-to-mock-a-base-class-with-python-mock-library

amikofalvy: There is a follow up ticket for writing tests.

amikofalvy (approved):
:shipit:

src/rev_ai/apiclient.py
JDongian: I personally think this change hurts readability.

https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

JDongian: at least for this line, with `a webhook` dangling below. I'm neutral with the changes to the other docstrings below.

PR #41, opened 2019-11-07
Feature/AISDK-70: Add custom vocab parameter to the streaming client Python

KostasRev:
Added parameter and Documentation for said parameter.

Do not currently have examples that submit custom vocabularies due to lack of custom vocabularies client. Awaiting [AISDK-69](https://github.com/revdotcom/revai-python-sdk/pull/40) to add these examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ streaming_client = RevAiStreamingClient("ACCESS TOKEN",
 `on_error`, `on_close`, and `on_connected` are optional parameters that are functions to be called when the websocket errors, closes, and connects respectively. The default `on_error` raises the error, `on_close` prints out the code and reason for closing, and `on_connected` prints out the job ID.
 If passing in custom functions, make sure you provide the right parameters. See the sample code for the parameters.
 
-Once you have a streaming client setup with a `MediaConfig` and access token, you can obtain a transcription generator of your audio. You can also use a custom vocabulary with your streaming job by supplying the custom vocabulary's ID when starting a connection!
+Once you have a streaming client setup with a `MediaConfig` and access token, you can obtain a transcription generator of your audio. You can also use a custom vocabulary with your streaming job by supplying the optional `custom_vocabulary_id` when starting a connection!
 
 ```python
 response_generator = streaming_client.start(AUDIO_GENERATOR, custom_vocabulary_id="CUSTOM VOCAB ID")
--- a/README.md
+++ b/README.md
@@ -157,10 +157,10 @@ streaming_client = RevAiStreamingClient("ACCESS TOKEN",
 `on_error`, `on_close`, and `on_connected` are optional parameters that are functions to be called when the websocket errors, closes, and connects respectively. The default `on_error` raises the error, `on_close` prints out the code and reason for closing, and `on_connected` prints out the job ID.
 If passing in custom functions, make sure you provide the right parameters. See the sample code for the parameters.
 
-Once you have a streaming client setup with a `MediaConfig` and access token, you can obtain a transcription generator of your audio.
+Once you have a streaming client setup with a `MediaConfig` and access token, you can obtain a transcription generator of your audio. You can also use a custom vocabulary with your streaming job by supplying the custom vocabulary's ID when starting a connection!
 
 ```python
-response_generator = streaming_client.start(AUDIO_GENERATOR)
+response_generator = streaming_client.start(AUDIO_GENERATOR, custom_vocabulary_id="CUSTOM VOCAB ID")
 ```
 
 `response_generator` is a generator object that yields the transcription results of the audio including partial and final transcriptions. The `start` method creates a thread sending audio pieces from the `AUDIO_GENERATOR` to our 
--- a/src/rev_ai/__init__.py
+++ b/src/rev_ai/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """Top-level package for rev_ai"""
 
-__version__ = '2.6.1'
+__version__ = '2.7.0'
 
 from .models import Job, JobStatus, Account, Transcript, MediaConfig, CaptionType
--- a/src/rev_ai/streamingclient.py
+++ b/src/rev_ai/streamingclient.py
@@ -76,14 +76,10 @@ class RevAiStreamingClient():
         })
 
         if custom_vocabulary_id:
-            url += '&' + urlencode({
-                'custom_vocabulary_id': custom_vocabulary_id
-                })
+            url += '&' + urlencode({'custom_vocabulary_id': custom_vocabulary_id})
 
         if metadata:
-            url += '&' + urlencode({
-                'metadata': metadata
-                })
+            url += '&' + urlencode({'metadata': metadata})
 
         try:
             self.client.connect(url)
--- a/src/rev_ai/streamingclient.py
+++ b/src/rev_ai/streamingclient.py
@@ -62,7 +62,7 @@ class RevAiStreamingClient():
         self.on_connected = on_connected
         self.client = websocket.WebSocket(enable_multithread=True)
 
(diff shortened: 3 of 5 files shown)


src/rev_ai/streamingclient.py
```diff
@@ -73,6 +73,9 @@ def start(self, generator, metadata=None):
             + '&content_type={}'.format(self.config.get_content_type_string()) \
             + '&user_agent={}'.format(quote('RevAi-PythonSDK/{}'.format(__version__), safe=''))
 
+        if custom_vocabulary_id:
```
JDongian: urlencode is the right way to create a URL. See https://github.com/revdotcom/revai-python-sdk/pull/32/files

src/rev_ai/streamingclient.py
```diff
@@ -73,6 +73,9 @@ def start(self, generator, metadata=None):
...
+            url += '&custom_vocabulary_id={}'.format(quote(custom_vocabulary_id), safe='')
```
JDongian: urlencode is the right way to create a URL. See https://github.com/revdotcom/revai-python-sdk/pull/32/files

amikofalvy:
In the future, if you are waiting on another PR to merge, you can base your new feature branch off your old feature branch. You don't need to wait for the first PR to land on develop.

src/rev_ai/streamingclient.py
amikofalvy: Josh, I asked Kyle about the status of your PR. I don't recall the status. Since we are modifying this directly, it makes sense to make these changes Kostas

README.md
```diff
@@ -157,10 +157,10 @@ streaming_client = RevAiStreamingClient("ACCESS TOKEN",
 `on_error`, `on_close`, and `on_connected` are optional parameters that are functions to be called when the websocket errors, closes, and connects respectively. The default `on_error` raises the error, `on_close` prints out the code and reason for closing, and `on_connected` prints out the job ID.
 If passing in custom functions, make sure you provide the right parameters. See the sample code for the parameters.
 
-Once you have a streaming client setup with a `MediaConfig` and access token, you can obtain a transcription generator of your audio.
+Once you have a streaming client setup with a `MediaConfig` and access token, you can obtain a transcription generator of your audio. You can also use a custom vocabulary with your streaming job by supplying the optional `custom_vocabulary_id` when starting a connection!
```
amikofalvy: I think any comment about CV is non sequitur at this location. The previous statement is about configuring the streaming client, CV is configured at request time.

src/rev_ai/__init__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """Top-level package for rev_ai"""
 
-__version__ = '2.6.1'
+__version__ = '2.7.0'
```
amikofalvy, line 4: 👍

README.md
amikofalvy, line 160: Exciting`!`

src/rev_ai/streamingclient.py
```diff
@@ -62,19 +62,24 @@ def __init__(self,
         self.on_connected = on_connected
         self.client = websocket.WebSocket(enable_multithread=True)
 
-    def start(self, generator, metadata=None):
+    def start(self, generator, custom_vocabulary_id=None, metadata=None):
```
amikofalvy: Ordering of parameters matters if the parameters are not named. You should add to the end of the parameter list otherwise this is a breaking change. @kbridbur Is this correct?

src/rev_ai/streamingclient.py
```diff
@@ -62,19 +62,24 @@ def __init__(self,
...
+            'content_type': self.config.get_content_type_string(),
+            'user_agent': 'RevAi-PythonSDK/{}'.format(__version__)
+        })
+
+        if custom_vocabulary_id:
+            url += '&' + urlencode({'custom_vocabulary_id': custom_vocabulary_id})
 
         if metadata:
-            url += '&metadata={}'.format(quote(metadata, safe=''))
+            url += '&' + urlencode({'metadata': metadata})
```
amikofalvy, line 82: Does this work with python 2.7 and python 3?

amikofalvy: I read this and it looks like urlencode is defined a little bit differently across python lang versions.
https://www.urlencoder.io/python/

src/rev_ai/streamingclient.py
kbridbur: You're correct, anyone who has
```
client.start(my_generator, my_metadata)
```
Will have broken code if we release this

src/rev_ai/streamingclient.py
kbridbur: It should work in both python 2 and 3. This is why we have the try catch import statement at the top

tests/test_streamingclient.py
```diff
@@ -100,3 +100,10 @@ def test_end(self, mock_streaming_client):
         mock_streaming_client.end()
 
         mock_streaming_client.client.abort.assert_called_once_with()
+
+
+def validate_query_parameters(called_url, query_dict):
```
kbridbur, line 105: This is fine here for now but we should move it to a utils file when we update all the client's to use `urlunparse`

tests/test_streamingclient.py
```diff
@@ -54,33 +54,33 @@ def test_constructor_no_token_no_config(self):
...
+        exp_responses = ['Connected, Job ID : testid\n',
+                         '{"type":"partial","transcript":"Test"}',
+                         'Connection Closed. Code : 1000; Reason : End of input. Closing\n']
         mock_streaming_client.client.recv_data.side_effect = data
 
-        response_gen = mock_streaming_client.start(mock_generator(), metadata)
+        response_gen = mock_streaming_client.start(mock_generator(), custom_vocabulary_id, metadata)
 
-        mock_streaming_client.client.connect.assert_called_once_with(url)
+        mock_streaming_client.client.connect.assert_called_once()
```
kbridbur: You should not use `assert_called_once()`. It was only introduced into unittest in python 3.6. It seems that previous versions of python don't catch that this is a problem because mocks in those versions (3.4 and 2.7) act as of you are just asking for a mock of some function called `assert_called_once`. You could probably confirm this by running 
```
what = mock_streaming_client.client.connect.assert_called_once()`
print(what)
```

See new in python 3.6: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once
Source: http://engineroom.trackmaven.com/blog/mocking-mistakes/

KostasRev: Ahh, makes sense. I'll resort to measuring the length of the calls list or some equivalent then.

src/rev_ai/streamingclient.py
KostasRev: You're right. Will change!

KostasRev:
@kbridbur or @amikofalvy Let me know if I can :shipit: 😄

kbridbur (approved):
:shipit:

src/rev_ai/streamingclient.py
```diff
@@ -62,19 +62,24 @@ def __init__(self,
...
         """Function to connect the websocket to the URL and start the response
             thread
 
         :param generator: generator object that yields binary audio data
         :param metadata: metadata to be attached to streaming job
         """
-        url = self.base_url + '?access_token={}'.format(self.access_token) \
-            + '&content_type={}'.format(self.config.get_content_type_string()) \
-            + '&user_agent={}'.format(quote('RevAi-PythonSDK/{}'.format(__version__), safe=''))
+        url = self.base_url + '?' + urlencode({
```
JDongian, line 72: it would have made more sense to minimize string manipulation and prepare the query parameters first, then only serialize once.

e.g.
```
params = {
...
}
if cond:
  params.update(...)
if cond2:
  params.update(...)

url = self.base_url + '?' + urlencode(params)
```

PR #43, opened 2020-01-09
AISDK-81: Use v1 of streaming api

kbridbur:


KostasRev (approved):
Looks good to me! I was not aware of setup.cfg prior to this.

kbridbur:
It's honestly not a totally necessary file. I've been considering deleting it, but we should keep it up to date as long as we have it

PR #44, opened 2020-01-29
AISDK-90: Add a Custom Vocabulary to the async example

KostasRev:


examples/async_example.py
```diff
@@ -15,13 +15,24 @@
...
 client = apiclient.RevAiAPIClient(access_token)
 
+# Construct CustomVocabulary objects using our premade class for ease
+mit_professors = CustomVocabulary(
+    ["Robert Berwick", "Noam Chomsky", "Evelina Fedorenko"]
+)
+# Or manually construct CustomVocabulary as such
+other_scientists = {"phrases": ["Albert Einstein"]}
+# Important: place custom vocabularies in a list before passing them into
+# either of client.submit_job_local_file() or sclient.ubmit_job_url()
```
nishchalb:
```suggestion
# either of client.submit_job_local_file() or client.submit_job_url()
```

examples/async_example.py
```diff
@@ -15,13 +15,24 @@
...
+custom_vocabularies = [mit_professors, other_scientists]
```
nishchalb, line 34: To me, it's unclear why these are passed in as a list. Is each individual vocabulary treated differently?

KostasRev: Our [API](https://www.rev.ai/docs#operation/SubmitTranscriptionJob) expects a list of CustomVocabulary instances. And an instance of CustomVocabulary contains a group of phrases. 

A CustomVocabulary instance can therefore contain contextually related custom phrases, and these may then be grouped together into a list in any desired combination to fit a user's needs for a certain transcription job.

examples/async_example.py
```diff
@@ -15,13 +15,24 @@
...
 
+# Construct CustomVocabulary objects using our premade class for ease
+mit_professors = CustomVocabulary(
+    ["Robert Berwick", "Noam Chomsky", "Evelina Fedorenko"]
+)
+# Or manually construct CustomVocabulary as such
+other_scientists = {"phrases": ["Albert Einstein"]}
+# Important: place custom vocabularies in a list before passing them into
+# either of client.submit_job_local_file() or client.submit_job_url()
+custom_vocabularies = [mit_professors, other_scientists]
```
kbridbur, line 34: We should use these in the job submission below

kbridbur (approved):
One minor comment, looks good otherwise

examples/async_example.py
amikofalvy: @nishchalb we made this design decision early on with the idea that we could support combining prebuilt custom-vocabularies in the future. At this point though, this does not appear to be a feature that customers want. This API interface gives us more flexibility in the future.

PR #45, opened 2020-03-02
AISDK-98: Add filter_profanity to the Python SDK

KostasRev:
Adding filter_profanity to both async and streaming clients.

### DIFF
--- a/examples/microphone_streaming_example.py
+++ b/examples/microphone_streaming_example.py
@@ -99,7 +99,7 @@ with MicrophoneStream(rate, chunk) as stream:
     # Uses try method to allow users to manually close the stream
     try:
         # Starts the server connection and thread sending microphone audio
-        response_gen = streamclient.start(stream.generator())
+        response_gen = streamclient.start(stream.generator(), filter_profanity=True)
 
         # Iterates through responses and prints them
         for response in response_gen:
--- a/src/rev_ai/apiclient.py
+++ b/src/rev_ai/apiclient.py
@@ -2,7 +2,7 @@
 """Speech recognition tools for using Rev.ai"""
 
 import json
-from .models import Job, Account, Transcript, CaptionType
+from .models import Account, CaptionType, Job, Transcript
 from .baseclient import BaseClient
 from . import utils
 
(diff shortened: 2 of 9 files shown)


src/rev_ai/apiclient.py
```diff
@@ -72,8 +73,15 @@ def submit_job_url(
             raise ValueError('media_url must be provided')
 
         payload = self._create_job_options_payload(
-            media_url, metadata, callback_url, skip_diarization,
-            skip_punctuation, speaker_channels_count, custom_vocabularies)
+            media_url,
```
amikofalvy: Use the pep 8 style guide: https://www.python.org/dev/peps/pep-0008/

amikofalvy (approved):
Make sure you are using the right python formatter. Otherwise, looks good.

src/rev_ai/streamingclient.py
```diff
@@ -81,6 +85,9 @@ def start(self, generator, metadata=None, custom_vocabulary_id=None):
         if metadata:
             url += '&' + urlencode({'metadata': metadata})
 
+        if filter_profanity:
+            url += '&' + urlencode({'filter_profanity': 'true'})
```
JohnDonavon, line 89: personal preference, but thoughts on using a format string here?

examples/microphone_streaming_example.py
```diff
@@ -99,7 +99,7 @@ def generator(self):
     # Uses try method to allow users to manually close the stream
     try:
         # Starts the server connection and thread sending microphone audio
-        response_gen = streamclient.start(stream.generator())
+        response_gen = streamclient.start(stream.generator(), filter_profanity=True)
```
JohnDonavon, line 102: Any reason why we have the example filter_profanity set to true here?

src/rev_ai/streamingclient.py
JohnDonavon: Also should we be defaulting this option to true?

JohnDonavon (commented):
A couple of comments/ questions.

examples/microphone_streaming_example.py
KostasRev: I'm just illustrating the usage as per the AC of the issue.

src/rev_ai/streamingclient.py
KostasRev: This is not defaulting to true. The default is None ([this line](https://github.com/revdotcom/revai-python-sdk/pull/45/files/29d2d44347d33919b3a0cfae7a1694a6a04f0a9d#diff-c79255447727fc4d20e17512f19774aeR69)).

As for format strings, if I recall correctly, usage of the library is more desirable as it handles any url formatting issues that might not be caught otherwise.

PR #48, opened 2020-03-15
AISDK-101: Fix Python Publish Jenkins Pipeline

amikofalvy:


Jenkinsfile
```diff
@@ -7,13 +7,19 @@ pipeline {
     }
     stages {
         stage('Build') {
+            agent {
+                docker {
+                    image "python:3"
```
amikofalvy: This allows us to use python 3's venv.

Jenkinsfile
```diff
@@ -7,13 +7,19 @@ pipeline {
...
             steps {
                 echo 'Building..'
                 sh '''
-                    virtualenv ./sdk-deploy
+                    python3 -m venv ./sdk-deploy
                     . ./sdk-deploy/bin/activate
-                    python setup.py sdist
-                    python setup.py bdist_wheel
+                    
+                    pip install wheel
```
amikofalvy: The subsequent command was failing unless this was installed.

Jenkinsfile
amikofalvy: @kbridbur

I want to make sure that the work being done in the build step is not being thrown away when we move on to the next step. which does not have the same docker agent specification.

Jenkinsfile
```diff
@@ -7,13 +7,19 @@ pipeline {
...
+            agent {
+                docker {
+                    image "python:3"
+                } 
+            }
             steps {
                 echo 'Building..'
                 sh '''
-                    virtualenv ./sdk-deploy
+                    python3 -m venv ./sdk-deploy
```
amikofalvy: Sets up a virtual environment using python 3

Jenkinsfile
```diff
@@ -7,13 +7,14 @@ pipeline {
     }
     stages {
         stage('Build') {
+            agent { docker { image "python:3" } }
             steps {
                 echo 'Building..'
                 sh '''
-                    virtualenv ./sdk-deploy
+                    rm -r dist
```
amikofalvy, line 14: Failed jobs don't remove the dist dir. This will force the directory to empty even if the previous job fails.

Jenkinsfile
```diff
@@ -27,6 +28,7 @@ pipeline {
             steps {
                 echo 'Testing..'
                 sh '''
+                    ls ./dist
```
amikofalvy: This was to double check on the directory contents between jobs. It's not harmful to include and it provides better clarity as to which files are included in the package directory.

Jenkinsfile
```diff
@@ -7,13 +7,14 @@ pipeline {
...
             steps {
                 echo 'Building..'
                 sh '''
-                    virtualenv ./sdk-deploy
+                    rm -r dist
+                    python3 -m venv ./sdk-deploy
                     . ./sdk-deploy/bin/activate
-                    python setup.py sdist
-                    python setup.py bdist_wheel
+                    python setup.py sdist bdist_wheel
```
kbridbur, line 17: Is there a reason you use python some places and python3 others?

Jenkinsfile
kbridbur: I'd personally rather not include stuff like this for debugging in the real deploy script. If someone is trying to debug they can replay it on ci and add these debug statements as they need

src/rev_ai/__init__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """Top-level package for rev_ai"""
 
-__version__ = '2.7.1'
+__version__ = '2.8.0'
```
kbridbur, line 4: Do we need this version bump. The SDK itself did not change

kbridbur (commented):
A couple questions

src/rev_ai/__init__.py
amikofalvy: This version change is not to note a version specific to these changes. It is to set the correct version for the upcoming python release. The version was set ahead of the release. Other additions to this release should get a minor version bump instead of a patch version bump.

Jenkinsfile
amikofalvy: Thanks. I had left one python3 accidentally from my testing.

I removed it since we are using a python 3 docker image.

kbridbur (approved):
:shipit:

PR #52, opened 2020-06-23
AISDK-108: Add delete endpoint for custom vocabulary

k-weng:


tests/test_customvocabulary.py
```diff
@@ -1,30 +1,72 @@
...
+from src.rev_ai.custom_vocabularies_client import RevAiCustomVocabulariesClient
 from src.rev_ai.models import CustomVocabulary
 
-phrases = ["Patrick Henry Winston", "Noam Chomsky"]
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
+
+TOKEN = "token"
```
amikofalvy, line 11:
```suggestion
ACCESS_TOKEN = "your_access_token"
```

tests/test_customvocabulary.py
```diff
@@ -1,30 +1,72 @@
...
+CV_ID = "cv1"
```
amikofalvy, line 12:
```suggestion
CV_ID = "cv12345678"
```

amikofalvy (approved):

examples/custom_vocabularies_example.py
```diff
@@ -41,11 +41,13 @@
...
     elif status == "failed":
         print("Job Failed : {}".format(custom_vocabularies["failure_detail"]))
         break
 
     if status == "complete":
         print("SUCCESS: {}".format(custom_vocabularies))
         break
+
+print("Deleting custom vocabulary")
+client.delete_custom_vocabulary(job_id)
```
amikofalvy: I would comment out this line and the previous.For someone who just runs the example of application and doesn't read every single line especially the last line in the file they would not understand why their custom vocabulary doesn't exist.

amikofalvy: To be clear I'd like to keep this line to illustrate how to use the delete endpoint. I just don't want it to execute

amikofalvy (approved):

examples/custom_vocabularies_example.py
k-weng: Makes sense!

PR #58, opened 2020-08-21
AISDK-128: Add delete_after_seconds option to streaming and async clients

KostasRev:


tests/test_streamingclient.py
```diff
@@ -107,8 +109,8 @@ def test_end(self, mock_streaming_client):
...
 
 
-def validate_query_parameters(called_url, query_dict):
+def validate_query_parameters(called_url, expected_query_dict):
     called_query_string = urlparse(called_url).query
     called_query_parameters = parse_qs(called_query_string)
-    for key in called_query_parameters:
-        assert called_query_parameters[key][0] == query_dict[key]
+    for key in expected_query_dict:
+        assert called_query_parameters[key][0] == expected_query_dict[key]
```
KostasRev, line 116: While modifying this test file it came to my attention that if `called_query_parameters` was somehow missing some expected values, the test would still pass. While fixing, it indeed became clear that the `called_url` that was being passed into `validate_query_parameters` had been the literal string "args". I believe the changes to line +89 now make it such that the `called_url` gets fetched properly.

src/rev_ai/apiclient.py
```diff
@@ -121,10 +122,9 @@ def submit_job_local_file(
             recognition to find those phrases. Each dictionary has the key
             "phrases" which maps to a list of strings, each of which represents
             a phrase you would like the speech recognition to bias itself toward.
-        :param filter_profanity: should rev.ai obscure profane words when
-                                 transcribing this file
-        :param remove_disfluencies: should rev.ai remove disfluencies when
-                                    transcribing this file
```
KostasRev, line 127: While writing docstrings for the streaming client I wrote a less verbose and more intuitive docstring for these two parameters and for the sake of consistency thought to copy over.

k-weng (approved):
:shipit:

PR #60, opened 2020-11-02
AISDK-133: Relax pinned version requirements to allow for easier integration of PythonSDK

### LINKED ISSUE #59: Pip version conflicts due to pinned requirements
Hi there, something I wanted to flag real quick because I'm bumping into it right now as a paying customer :-)

The `revai-python-sdk` library specifies exact (`==`) version matches in it's `setup.py` dependencies via `requirements.txt`:

```
requests==2.21.0
enum34==1.1.6
six==1.12.0
websocket-client==0.56.0
mock==3.0.5
```

([Source](https://github.com/revdotcom/revai-python-sdk/blob/f1196b88e8ff206d92a3fe91c02aae93c6c266ab/requirements.txt))

Since many other libraries depend on commonly used dependencies like `six` it's easy to run into version conflicts in larger projects with `six` (and other dependencies) pinned to exact version numbers with the `==` version specifier. 

It would be preferable if the dependencies for `revai-python-sdk` could be made more flexible.

For example, see how Google handle this in their Python API SDK: https://github.com/googleapis/google-api-python-client/blob/a527de24cda7c7d62a371203158fd5f617a1c08c/setup.py#L37-L44

(Also, it looks like `mock` should be moved to `requirements_dev.txt` since it's only used in the tests as far as I can tell.)

Thanks for looking into it!

KostasRev:


k-weng:
Does this change require a version update? And can you link the issue to this PR so it will automatically close once this PR is merged?

src/rev_ai/__init__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """Top-level package for rev_ai"""
 
-__version__ = '2.10.0'
+__version__ = '2.10.1'
```
KostasRev, line 4: @k-weng looking at [semver spec](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) I am thinking of bumping the PATCH version since we are not adding functionality but are introducing a change.

k-weng (approved):
LGTM

requirements.txt
```diff
@@ -1,5 +1,4 @@
-requests==2.21.0
-enum34==1.1.6
-six==1.12.0
-websocket-client==0.56.0
-mock==3.0.5
\ No newline at end of file
+requests>=2.21.0
```
amikofalvy: Would this work with requests 3.0?

amikofalvy: We wouldn't necessarily want it to work with request 3.0 but I imagine we would allow requests 2.x.x

KostasRev: I think it makes sense to restrict the range to be `requests>=2.21.0, requests<3.0.0` and same for the other deps.

amikofalvy (approved):

requirements.txt
```diff
@@ -1,5 +1,4 @@
-requests==2.21.0
-enum34==1.1.6
-six==1.12.0
-websocket-client==0.56.0
-mock==3.0.5
\ No newline at end of file
+requests>=2.21.0<3.0.0
```
k-weng:
```suggestion
requests>=2.21.0,<3.0.0
```

PR #63, opened 2020-12-22
AISDK-136: Add Language parameter to the Python SDK

lrgottlieb:


README.md
```diff
@@ -159,7 +159,7 @@ If passing in custom functions, make sure you provide the right parameters. See
 
 Once you have a streaming client setup with a `MediaConfig` and access token, you can obtain a transcription generator of your audio. You can also use a custom vocabulary with your streaming job by supplying the optional `custom_vocabulary_id` when starting a connection!
 
-More optional parameters can be supplied when starting a connection, these are `metadata`, `filter_profanity`, `remove_disfluencies`, and `delete_after_seconds`. For a description of these optional parameters look at our [streaming documentation](https://www.rev.ai/docs/streaming#section/WebSocket-Endpoint).
+More optional parameters can be supplied when starting a connection, these are `metadata`, `filter_profanity`, `remove_disfluencies`, `delete_after_seconds` and `language`. For a description of these optional parameters look at our [streaming documentation](https://www.rev.ai/docs/streaming#section/WebSocket-Endpoint).
```
amikofalvy: This text block is for streaming api. Language is not supported for streaming api. This line should be reverted.

examples/async_example.py
```diff
@@ -56,6 +57,7 @@
                             custom_vocabularies=custom_vocabularies,
                             filter_profanity=False,
                             remove_disfluencies=False,
+                            language=None,
```
amikofalvy: I think new parameters should be appended to the end of the parameters list after `delete_after_seconds`. It is not a requirement for customers to name their parameters when invoking these functions. Inserting a parameter before a previously existing parameter is a breaking change to the function interface.

amikofalvy: Please implement the change here and elsewhere in this PR.

HISTORY.rst
```diff
@@ -67,4 +67,9 @@ History
...
 * Add custom vocabulary to async example
 * Add filter_profanity to async and streaming clients, examples, and documentation
-* Add remove_disfluencies to async client
\ No newline at end of file
+* Add remove_disfluencies to async client
+
+2.7.2
+------------------
+
+* Add language to async client
```
amikofalvy: We should clarify that this is the `language` job option. In the previous changelog notes it is more obvious that these (remove_disfluencies, fitler_orifanity) are features of the client, but `language` is too generic a term for someone to infer that this is the job property and not just new wording of the async client.

tests/test_job.py
```diff
@@ -136,6 +138,7 @@ def test_submit_job_url_with_success(self, mock_session, make_mock_response):
                 'custom_vocabularies': CUSTOM_VOCAB,
                 'filter_profanity': True,
                 'remove_disfluencies': True,
+                'language' = LANG,
```
amikofalvy: I expect this test will fail since the  action of the test does not include language parameter in the request. 
Line 119:
```
res = client.submit_job_url(MEDIA_URL, METADATA,
                                    CALLBACK_URL, True,
                                    True, 1, CUSTOM_VOCAB, True, True, 0)
                               ```

Comment (amikofalvy) [reply tests/test_job.py]: You should modify the quoted line to also pass LANG.

File: tests/test_job.py [hunk 5]
```diff
@@ -21,6 +21,7 @@
 JOB_ID_URL = urljoin(RevAiAPIClient.base_url, 'jobs/{}'.format(JOB_ID))
 JOBS_URL = urljoin(RevAiAPIClient.base_url, 'jobs')
 CUSTOM_VOCAB = [{"phrases": ["word one", "word two"]}]
+LANG = 'en'
```
amikofalvy: relatively minor feedback - In general I encourage devs to not use abbreviations. It helps make variable names exceptionally clear
```suggestion
LANGUAGE = 'en'
```

tests/test_job.py
```diff
@@ -108,6 +109,7 @@ def test_submit_job_url_with_success(self, mock_session, make_mock_response):
             'speaker_channels_count': 1,
             'filter_profanity': True,
             'remove_disfluencies': True,
+            'language': LANG,
```
amikofalvy: it makes no difference in execution but I would order these to match the order of the parameters with language coming last. Make this change here and below and in the submit local file unit test as well.

tests/test_job.py
```diff
@@ -158,6 +161,7 @@ def test_submit_job_local_file_with_success(self, mocker, mock_session, make_moc
             'speaker_channels_count': 1,
             'filter_profanity': True,
             'remove_disfluencies': True,
+            'language' = LANG,
             'delete_after_seconds': 0
         }
         response = make_mock_response(url=JOB_ID_URL, json_data=data)
```
amikofalvy, line 168: You will need to modify a few lines lower to pass LANG to `client.submit_job_local_file(FILENAME, METADATA, CALLBACK_URL, True,True, 1, CUSTOM_VOCAB, True, True, 0, LANGUAGE)`

amikofalvy (changes requested):

tests/test_job.py
```diff
@@ -136,7 +138,8 @@ def test_submit_job_url_with_success(self, mock_session, make_mock_response):
                 'custom_vocabularies': CUSTOM_VOCAB,
                 'filter_profanity': True,
                 'remove_disfluencies': True,
-                'delete_after_seconds': 0
+                'delete_after_seconds': 0,
+                'language' = LANG
```
amikofalvy:
```suggestion
                'language': LANGUAGE
```

amikofalvy: I noticed this when I checked the CI build of your commit: https://travis-ci.org/github/revdotcom/revai-python-sdk/jobs/751115476

tests/test_job.py
```diff
@@ -158,7 +161,8 @@ def test_submit_job_local_file_with_success(self, mocker, mock_session, make_moc
             'speaker_channels_count': 1,
             'filter_profanity': True,
             'remove_disfluencies': True,
-            'delete_after_seconds': 0
+            'delete_after_seconds': 0,
+            'language' = LANG
```
amikofalvy:
```suggestion
            'language': LANG
```

amikofalvy (commented):
You should be able to run the test suite locally to ensure that your changes are valid syntax.

I haven't tested them myself, but I think you can create a virtual environment and create a test suite with these commands: 
https://github.com/revdotcom/revai-python-sdk/blob/develop/Jenkinsfile#L30-L34

tests/test_job.py
lrgottlieb: Oh that was a silly error.

amikofalvy (commented):
The changes look good. I want to wait for CI to pass before giving the final :shipit:

amikofalvy (approved):
Can you add a new section to the README.md which includes instructions on how to run the test suite and linter locally?

HISTORY.rst
```diff
@@ -67,4 +67,9 @@ History
 * Use v1 of the streaming api
 * Add custom vocabulary to async example
 * Add filter_profanity to async and streaming clients, examples, and documentation
-* Add remove_disfluencies to async client
\ No newline at end of file
+* Add remove_disfluencies to async client
+
+2.7.2
```
amikofalvy: The next version is actually 2.11.0

amikofalvy (changes requested):
We need to increase the package version in setup.cfg

amikofalvy (approved):

PR #66, opened 2021-02-04
feature/AISDK-139. Additions to allow use of custom_vocabulary_id

lrgottlieb:


src/rev_ai/apiclient.py
```diff
@@ -130,8 +133,9 @@ def submit_job_local_file(
         :param filter_profanity: whether to mask profane words
         :param remove_disfluencies: whether to exclude filler words like "uh"
         :param delete_after_seconds: number of seconds after job completion when job is auto-deleted
-        :param language: specify language using ISO 639-1 2-letter language
-                         code
+        :param language: specify language using ISO 639-1 2-letter language code
+        :param custom_vocabulary_id: The id of a pre-completed custom vocabulary
+            submitted through the custom vocabularies. Exclusive with custom_vocabularies.
```
KostasRev: Do you think it might be clearer to say "Cannot be used with custom_vocabularies param"?

Either way, I think we should also put the equivalent statement of exclusivity on `custom_vocabularies`.

lrgottlieb: I've updated the language here.

KostasRev (approved):
Please fix lint error and merge.

PR #68, opened 2021-06-09
AISDK-157: Change language description in python in-line doc should reference API Reference

beaudrychase:


src/rev_ai/apiclient.py
```diff
@@ -73,7 +73,8 @@ def submit_job_url(
         :param filter_profanity: whether to mask profane words
         :param remove_disfluencies: whether to exclude filler words like "uh"
         :param delete_after_seconds: number of seconds after job completion when job is auto-deleted
-        :param language: specify language using ISO 639-1 2-letter language code
+        :param language: specify language using ISO 639-1 2-letter language code with the exception 
+            of Mandarin ("cmn") which uses a ISO 639-3 language code
```
amikofalvy: I think this is too specific. I would like to "future proof" this documentation so that we don't have to update it when we add new languages.

I think mentioning the ISO specs is reasonable, just remove reference to any specific language.

src/rev_ai/apiclient.py
```diff
@@ -135,10 +135,11 @@ def submit_job_local_file(
...
```
amikofalvy: ditto here

amikofalvy (changes requested):
Please remove specific language references to make the documentation less fragile

src/rev_ai/apiclient.py
beaudrychase: would "with the exception of languages with no ISO 639-1 2-letter language code" be better then?

amikofalvy: How about this?

```suggestion
        :param language: specify language using the one of the supported ISO 639-1 (2-letter) or ISO 639-3 (3-letter) language codes as defined in the API Reference
```

beaudrychase: That is much better

amikofalvy (approved):

hbkse (approved):

PR #69, opened 2021-06-30
Added detailed_partials parameter to start streaming method

vhrustic:
In documentation we have `detailed_partials` parameter, which is in beta, but should be available in SDK:
https://www.rev.ai/docs/streaming#section/WebSocket-Endpoint/Detailed-Partials-(Beta)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.12.1
+current_version = 2.13.0
 commit = True
 tag = True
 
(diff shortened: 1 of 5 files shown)


amikofalvy:
@vhrustic. Thanks for opening a PR. We are going to expand on your work to add support for detailed partials to the python SDK. We hadn't added support for this feature previously because it was in beta.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.12.0
+current_version = 2.12.1
```
amikofalvy: I think you should bump the minor version since this is additive functionality and not a minor bug fix.

```suggestion
current_version = 2.13.0
```

amikofalvy (approved):
One comment on symver number, otherwise looks good. Nice test improvements.

PR #73, opened 2022-01-27
AISDK-154: Add support for human transcription

k-weng:


src/rev_ai/models/asynchronous/job.py
```diff
@@ -14,7 +14,16 @@ def __init__(
...
+            duration_seconds=None,
+            delete_after_seconds=None,
+            skip_diarization=None,
+            skip_punctuation=None,
+            remove_disfluencies=None,
+            filter_profanity=None,
+            custom_vocabulary_id=None,
+            speakers_channel_count=None,
+            language=None,
+            transcriber=None):
```
k-weng: Added missing properties to job model

README.md
```diff
@@ -54,6 +54,23 @@ If you want to get fancy, both send job methods take `metadata`, `callback_url`,
...
+Check out our documentation for [Human Transcription](https://www.rev.ai/docs#section/Human-Transcription-(Labs)) for more details.
+
+```python
+# submitting a human transcription jobs
+job = client.submit_job_url("https://example.com/file-to-transcribe.mp3",
+    transcriber='human',
+    verbatim=False,
+    rush=False,
+    test_mode=True
+    segments_to_transcribe=[{
```
kbridbur, line 68: What do you think of having a segment object instead of this being a list of dictionaries

kbridbur (commented):
One question

README.md
hbkse: I **_think_** the pattern in python is not to use objects but to just rely on type hints in `apiclient.py`, which I guess we don't do either

amikofalvy: @gweng88 can you give an example or link to type hints?

amikofalvy (approved):

README.md
hbkse: I did a bit of research and the decision to use objects and the decision to use type hints seem pretty separate. Here's some of the things I looked at: 
- [Google's python style guide note on type hints](https://google.github.io/styleguide/pyguide.html#221-type-annotated-code)
- [AWS boto3 uses dicts for most user facing functions](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.put_item) when you could argue they could use objects.
- [SDKs generated from openapi specs](https://github.com/plaid/plaid-python/blob/master/plaid/model/item_public_token_exchange_request.py#L28) expose objects to users (this is probably the only way the generation can work)

kbridbur: Seems reasonable, lets leave it as is

kbridbur (approved):

PR #78, opened 2022-04-01
AISDK-175: Update Python SDK with Rev.ai to Rev AI

kshiflett88:
One question I had about this [line](https://github.com/revdotcom/revai-python-sdk/blob/83b67b6474ac0a2aea2bb13680e0afede19d6d63/src/rev_ai/apiclient.py#L65) (there are four similar lines like this in this file) with it being lower case and the context sounds more like it was specific to the url rev.ai and not the over all brand Rev AI. I may be reading wrong just needed clarification if this line and the others similar in the file should be changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Rev.ai Python SDK
+# Rev AI Python SDK
 
 [![CI](https://github.com/revdotcom/revai-python-sdk/actions/workflows/build_test.yml/badge.svg)](https://github.com/revdotcom/revai-python-sdk/actions/workflows/build_test.yml)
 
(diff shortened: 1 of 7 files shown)


amikofalvy:
> One question I had about this [line](https://github.com/revdotcom/revai-python-sdk/blob/83b67b6474ac0a2aea2bb13680e0afede19d6d63/src/rev_ai/apiclient.py#L65) (there are four similar lines like this in this file) with it being lower case and the context sounds more like it was specific to the url rev.ai and not the over all brand Rev AI. I may be reading wrong just needed clarification if this line and the others similar in the file should be changed.

I interpret this as the service name. The service name is "Rev AI". Please update these references from `rev.ai` to `Rev AI`.

amikofalvy (approved):

PR #79, opened 2022-04-04
AISDK-187: Add generic api client and topic extraction implementation

kbridbur:
TODO:
[x] - Tests
[x] - Clean up example
[x] - Bump version


src/rev_ai/generic_api_client.py
```diff
@@ -0,0 +1,161 @@
+from .baseclient import BaseClient
+
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
+
+
+class GenericApiClient(BaseClient):
```
kbridbur, line 12: This is the main piece that needs feedback. I'm currently unhappy with how submit is functioning. I could use **kwargs and just create a dictionary from all of that but then users don't get argument hints or named arguments

src/rev_ai/models/basejob.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+"""Base Job model"""
+
+from .asynchronous.job_status import JobStatus
+
+
+class BaseJob:
```
kbridbur: I created this class for insights job and async job to inherit but there doesn't seem to be any actual benefit to this. I'll probably remove it

src/rev_ai/topic_extraction_client.py
```diff
@@ -0,0 +1,50 @@
+from .generic_api_client import GenericApiClient
+from .models import TopicExtractionJob, TopicExtractionResult
+
+
+class TopicExtractionClient(GenericApiClient):
```
kbridbur, line 8: I'm pretty happy with this class. I think it looks very clean, basically all the functionality is in the GenericApiClient and it works perfectly

examples/topic_example.py
```diff
@@ -0,0 +1,69 @@
...
+api_client = apiclient.RevAiAPIClient(access_token)
+
+async_job_id = ""
+
+transcript = api_client.get_transcript_object(async_job_id)
+job = client.submit_job(json=transcript)
+
+print("Submitted Job")
+
+while True:
```
beaudrychase: Should we add a note that this flow isn't recommended for production and that using a callback url is recommended

src/rev_ai/generic_api_client.py
beaudrychase: I'm a bit confused by your comment. Isn't the goal to not have users call `_submit` directly and instead use the endpoint specific wrappers? Wouldn't it not matter that there aren't argument hints for `_submit`? I think it would be good to use **kargs as the only input to submit

beaudrychase (commented):
left a few comments

src/rev_ai/generic_api_client.py
```diff
@@ -0,0 +1,161 @@
...
+                             account. Generated on the settings page of your account dashboard
+                             on Rev AI.
+        :param api_name: name of the api to submit to
+        :param api_version: version of the api to submit to
+        :param parse_job_info: method to be used to parse job information
+        :param parse_job_result: method to be used to parse job results
+        """
+
+        BaseClient.__init__(self, access_token)
+        self.base_url = 'https://api.rev.ai/{0}/{1}/'.format(api_name, api_version)
```
kbridbur, line 29: I don't see any reason to

examples/topic_example.py
kbridbur: Seems reasonable

src/rev_ai/generic_api_client.py
kbridbur: Users shouldn't be using this class directly at all. They should be using one of the superclasses which we will provide them TopicExtractionClient is an example of one such class. 

To better explain why I'm unhappy with the current solution: if you look at TopicExtraction client it only needs to have 1 method which is to create its own submit method with explicit "text" and "json" parameters. All the other methods from GenericApiClient just work without TopicExtractionClient having to do anything at all. It would be nice if submit worked in the same way but I don't think its possible to have both that and keep with Python argument conventions.

In our other SDKs I imagine this generic client will just also take a "Submission" object which contains all the information needed to submit a job and they won't need to know the internals of that. However in Python convention is to not have these submission objects and instead have individual arguments on the function directly. An example of this can be seen in the difference between our async api submit calls for [python](https://github.com/revdotcom/revai-python-sdk/blob/develop/src/rev_ai/apiclient.py#L39) and [node](https://github.com/revdotcom/revai-node-sdk/blob/develop/src/api-client.ts#L93). Node takes an options object while python explicitly lists out every argument

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
+"""Copyright 2019 REV
```
amikofalvy, line 1: Do you need to add LICENSE to these example files? It can be difficult to maintain them all. Can we just rely on the top level LICENSE?

src/rev_ai/generic_api_client.py
```diff
@@ -0,0 +1,161 @@
...
+
+        query = '?{}'.format('&'.join(params))
+        response = self._make_http_request(
+            "GET",
+            urljoin(self.base_url, 'jobs{}'.format(query))
+        )
+
+        return [self.parse_job_info(job) for job in response.json()]
+
+    def get_result_object(self, id_):
```
beaudrychase: This will need a kwargs to support things like `threshold` for topic extraction and `filter_for` for sentiment analysis

examples/topic_extraction_example.py
kbridbur: I'm not really sure. Maybe @vikram-rev would know

amikofalvy: For all licensing questions, I refer to legal.

kbridbur: Who on legal do you normally ask?

src/rev_ai/generic_api_client.py
```diff
@@ -0,0 +1,167 @@
...
+        :param limit: optional, limits the number of jobs returned,
+                      if none, a default of 100 jobs is returned, max limit if 1000
+        :param starting_after: optional, returns jobs created after the job with this id,
+                               exclusive (job with this id is not included)
+        :returns: list of jobs response data
+        :raises: HTTPError
+        """
+        params = []
+        if limit is not None:
+            params.append('limit={}'.format(limit))
```
beaudrychase, line 81:
```suggestion
            params.append(f'limit={limit}')
```

beaudrychase: I think pretty much everywhere you use `.format` it should be replaced with this pattern

examples/topic_extraction_example.py
vikram-rev: AFAIK it's generally recommended to add a license to all files and also at the top-level. There are tools that do it for you automatically, including in VScode. See https://opensource.google/documentation/reference/releasing/preparing#license-headers for a more detailed discussion. +1 to @amikofalvy suggestion to check with a lawyer as first preference though.

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if status == "COMPLETED":
+        # Getting a list of current topic extraction jobs connected with your account
+        # The optional parameters limits the length of the list.
+        # starting_after is a job id which causes the removal of
+        # all jobs from the list which were created before that job
+        list_of_jobs = client.get_list_of_jobs(limit=None, starting_after=None)
+
+        # obtain a list of topics and their scores for the job
+        result = client.get_result_object(job.id)
```
beaudrychase:
```suggestion
        result = client.get_result_object(job.id, threshold=None)
```
Do we want to mention the threshold option?

src/rev_ai/generic_api_client.py
kbridbur: fstrings are only available for python3+. We want to continue supporting python2 for the time being

src/rev_ai/generic_api_client.py
```diff
@@ -0,0 +1,167 @@
...
+                             a webhook
+        :param delete_after_seconds: number of seconds after job completion when job is auto-deleted
+        :param language: specify language using the one of the supported ISO 639-1 (2-letter) or
+            ISO 639-3 (3-letter) language codes as defined in the API Reference
+        :returns: Job info object
+        :raises: HTTPError
+        """
+        payload = {}
+        for key, value in kwargs.items():
+            if value:
```
hbkse: won't this not work if the values are intentionally false-y, like `skip_punctuation=false` 
```suggestion
            if value is not None:
```

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import time
+
```
dmtrrk: do we need this line? and there are 2 lines empty after imports

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+from rev_ai import topic_extraction_client
+
+
+# helper function for printing nice results, can be ignored
```
dmtrrk: Please update the commenting style as per [PEPs](https://peps.python.org/pep-0257/)

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+def remove_none_elements(dictionary):
```
dmtrrk: Just another option:
We can replace this function with the local lambda:

```
remove_none_elements = lambda dictionary: {key: value for key, value in dictionary.items() if value}
```

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+    none_less = {}
+    for key in dictionary:
+        if dictionary[key]:
+            none_less[key] = dictionary[key]
+    return none_less
+
+
+# String containing your access token
+access_token = "your_access_token"
```
dmtrrk: we had a discussion to use `<>` for placeholders
cc: @amikofalvy

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+# uncommenting the lines below
+#
+# async_job_id = "your_job_id"
+# async_api_client = apiclient.RevAiAPIClient(access_token)
+# transcript = api_client.get_transcript_object(async_job_id)
+# transcript_json = transcript
+# text = None
+
+# Submitting a job with either text or json. Only one can be provided per submission
+job = client.submit_job(text=text, json=transcript_json)
```
dmtrrk: can it be more clear to have 2 functions `submit_from_text`, `submit_from_transcript`?

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+print("Submitted Job")
+
+while True:
+
+    # Obtains details of a job in json format
+    job_details = client.get_job_details(job.id)
```
dmtrrk, line 51: we don't have `get job details` API, but [GetJobById](https://docs.rev.ai/api/topic-extraction/reference/#operation/GetTopicExtractionJobById)

src/rev_ai/generic_api_client.py
dmtrrk: My concern here is that trying to pass parameters and make it more Python-ish, we are making it [less Python](https://peps.python.org/pep-0020/) (e.g. `Explicit is better than implicit.`).
Once `_submit_job` does not know what payload parameters should be there, maybe it is not its responsibility? So maybe we can assume that the payload is a single argument and `def _submit_job(self, payload)`?

kbridbur: Interesting. I'd be happy to discuss this at length. It seems like a choice between being more explicit and the code being DRY. If I do as you're suggesting well have to duplicate payload construction logic everywhere

src/rev_ai/generic_api_client.py
kbridbur: Good catch. I'll do an explicit None check

src/rev_ai/generic_api_client.py
dmtrrk: we can tell this:
1. `_submit_xx` is responsible to submit the provided payload
2. `submit_xx` is responsible to construct a concrete payload and submit it using the `_submit`
3. `_enhance_payload(payload, callback, ...)` is called from `submit_`

```
def _submit(payload):
  http.submit(payload)

def _enhance(payload, callback):
  if callback:
    payload['callback'] = callback
 # or return a new dict

def submit_topic(payload):
  payload = {}
  self.enhance(payload, 'url')
  self._submit(payload)
```

examples/topic_extraction_example.py
kbridbur: I named it this to be consistent with the method name in the async client. I could change both but that would require a major version bump which seems unnecessary

examples/topic_extraction_example.py
kbridbur: I've updated the topic extraction example. Please confirm that this matches what you expect and I'll fix the other examples

examples/topic_extraction_example.py
kbridbur: Sure, that seems like it fits python style better anyways

examples/topic_extraction_example.py
amikofalvy: @kbridbur Please refer to our [Style Guide](https://github.com/revdotcom/revai-docs/blob/production/guidelines.md)

src/rev_ai/generic_api_client.py
kbridbur: Discussed offline. We will go with this approach as it is more readable and better follows python standards of being explicit even though it slightly increases complexity and code reuse

beaudrychase (approved):
Looks good to me

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+# job = client.submit_job_from_transcript(transcript_json,
+#                                         metadata=None,
+#                                         callback_url=None,
+#                                         delete_after_seconds=None,
+#                                         language=None)
+
+print("Submitted Job")
+
+while True:
+
```
dmtrrk: is there a reason to have so many newlines? double-newlines above

examples/topic_extraction_example.py
```diff
@@ -0,0 +1,98 @@
...
+    if status == "COMPLETED":
+        # Getting a list of current topic extraction jobs connected with your account
+        # The optional parameters limits the length of the list.
+        # starting_after is a job id which causes the removal of
+        # all jobs from the list which were created before that job
+        list_of_jobs = client.get_list_of_jobs(limit=None, starting_after=None)
+
+        # obtain a list of topics and their scores for the job
+        result = client.get_result_object(job.id, threshold=None)
+        remove_none_elements = lambda dictionary: {
```
dmtrrk:
```suggestion
        remove_none_elements = lambda dictionary: {k: v for k, v in dictionary.items() if v}
```

dmtrrk: not sure whether it is better. better 2-liner looks unusual

```
remove_none_elements = lambda dictionary: {
  key: value
    for key, valu in dictionary.items() if v}
```

src/rev_ai/generic_api_client.py
```diff
@@ -0,0 +1,152 @@
...
+
+        self._make_http_request(
+            "DELETE",
+            urljoin(self.base_url, 'jobs/{}'.format(id_)),
+        )
+
+        return
+
+    def _enhance_payload(self, payload, metadata, callback_url, delete_after_seconds):
+        enhanced = payload
```
dmtrrk:
```suggestion
        enhanced = payload.copy()
```

src/rev_ai/generic_api_client.py
```diff
@@ -0,0 +1,152 @@
...
+            raise ValueError('id_ must be provided')
+
+        self._make_http_request(
+            "DELETE",
+            urljoin(self.base_url, 'jobs/{}'.format(id_)),
+        )
+
+        return
+
+    def _enhance_payload(self, payload, metadata, callback_url, delete_after_seconds):
```
dmtrrk, line 147:
```suggestion
    def _create_payload(self, payload, metadata, callback_url, delete_after_seconds):
```

src/rev_ai/models/insights/topic_extraction/topic_extraction_result.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+"""Topic extraction result model"""
+
```
dmtrrk, line 3: sometime it is a [single line break](https://github.com/revdotcom/revai-python-sdk/pull/79/files#diff-6cb75734fa52c453b1067020e064b4a351b5dd91bbe1177d2e2be677357d1f63R3), sometimes double

kbridbur: It is a single line break between top of file and imports and a double between top of file and class declarations. As far as I can tell this is consistent

examples/topic_extraction_example.py
kbridbur: I removed some that seemed unnecessary

PR #80, opened 2022-04-04
AISDK-184: Implement Link Checker in Python SDK

kshiflett88:


.github/workflows/.lycheeignore
```diff
@@ -0,0 +1,5 @@
+# API Domain stubs used as url partials in the api clients
```
amikofalvy: I think the ignore file is not working because it is not at the top level of the repo. If you move the `.lycheeignore` file to the top level then the link checker should start to pass.

amikofalvy (changes requested):
Please move lycheeignore file

k-weng (approved):

amikofalvy (approved):

PR #83, opened 2022-04-18
AISDK-190 Update sdks to support notification_config and source_config

aaron-wilson-rev:


README.md
```diff
@@ -50,7 +50,7 @@ job = client.submit_job_url("https://example.com/file-to-transcribe.mp3")
 `job` will contain all the information normally found in a successful response from our
 [Submit Job](https://docs.rev.ai/api/asynchronous/reference/#operation/SubmitTranscriptionJob) endpoint.
 
-If you want to get fancy, both send job methods take `metadata`, `callback_url`,
+If you want to get fancy, both send job methods take `metadata`, `notification_url`, `notification_auth`,
```
amikofalvy: Can we update this to use the same term `submit` instead of `send`?

```suggestion
If you want to get fancy, both submit job methods take `metadata`, `notification_url`, `notification_auth`,
```

examples/async_example.py
```diff
@@ -53,7 +54,8 @@
 url = "https://www.rev.ai/FTC_Sample_1.mp3"
 job = client.submit_job_url(url,
                             metadata=None,
-                            callback_url=None,
```
amikofalvy, line 56: I thought we discussed leaving the callback_url option as a parameter even if it is deprecated. If you remove this parameter then we will have to update the version to indicate a breaking change and it could be an inconvenience to our customers.

src/rev_ai/apiclient.py
```diff
@@ -438,18 +449,19 @@ def _create_job_options_payload(
             test_mode=None,
             segments_to_transcribe=None):
         payload = {}
-        if media_url:
-            payload['media_url'] = media_url
+        if source_url:
+            payload['source_config'] = CustomerUrlData(source_url, source_auth).to_dict()
```
amikofalvy: Since the source url and auth are tied together why not make the user pass the object into `submit_job_url` as a parameters `notification_config` and `source_config`?

amikofalvy (changes requested):
You already have configured the CustomerUrlData class, why not show users that auth and url are coupled by making the user create an instance of CustomerUrlData and passing it in as either notification config or source config?

examples/async_example.py
aaron-wilson-rev: I thought we would be updating the version regardless, so they would only have to change if they updated. But it makes more sense to keep it as deprecated

src/rev_ai/apiclient.py
aaron-wilson-rev: My thought was it might be simpler to abstract away the structure of the request, but since those options are very much dependent on each other it might be better to show that

src/rev_ai/apiclient.py
```diff
@@ -44,9 +45,11 @@ def __init__(self, access_token):
 
     def submit_job_url(
             self,
-            media_url,
+            source_url,
+            source_auth=None,
```
aaron-wilson-rev: @amikofalvy A lot of my confusion is around this, where we have a media_url parameter that's required while nothing else is, and we want to add the source config option, which is not required, but also replaces the need for the media url. 

Can we make media_url an optional parameter without breaking implementations?  Can hop on a zoom call if you want.

aaron-wilson-rev: I added a default option for media_url and added source_config at the end, so if anyone is relying on order of args rather than named args it won't break anything.

README.md
```diff
@@ -50,9 +50,17 @@ job = client.submit_job_url("https://example.com/file-to-transcribe.mp3")
 `job` will contain all the information normally found in a successful response from our
 [Submit Job](https://docs.rev.ai/api/asynchronous/reference/#operation/SubmitTranscriptionJob) endpoint.
 
-If you want to get fancy, both send job methods take `metadata`, `callback_url`,
-`skip_diarization`, `skip_punctuation`, `speaker_channels_count`, `custom_vocabularies`, `filter_profanity`, `remove_disfluencies`, `delete_after_seconds`, `language`, and `custom_vocabulary_id` as optional parameters, these are described in the request body of
-the [Submit Job](https://docs.rev.ai/api/asynchronous/reference/#operation/SubmitTranscriptionJob) endpoint.
+If you want to get fancy, both submit job methods take `metadata`,`notification_config`,
```
amikofalvy, line 53: missing source_config

src/rev_ai/apiclient.py
```diff
@@ -102,14 +102,27 @@ def submit_job_url(
...
         :raises: HTTPError
         """
-        if not media_url:
-            raise ValueError('media_url must be provided')
-        payload = self._create_job_options_payload(media_url, metadata,
-                                                   callback_url, skip_diarization,
-                                                   skip_punctuation, speaker_channels_count,
+        if not (media_url or source_config):
+            raise ValueError('media_url or source_config must be provided')
+        check_exclusive_options(media_url, 'media_url', source_config, 'source_config')
```
amikofalvy: We should rely on the API to validate the model. Don't make extra work for yourself by performing the model validation client side.

src/rev_ai/apiclient.py
```diff
@@ -143,18 +156,16 @@ def submit_job_local_file(
...
 
         :param filename: path to a local file on disk
         :param metadata: info to associate with the transcription job
-        :param callback_url: callback url to invoke on job completion as a
-                             webhook
-        :param skip_diarization: should Rev AI skip diaization when
-                                 transcribing this file
-        :param skip_punctuation: should Rev AI skip punctuation when
-                                 transcribing this file
+        :param callback_url: callback url to invoke on job completion as a webhook
```
amikofalvy, line 159: We should note that this parameter is deprecated.

https://stackoverflow.com/questions/63902170/mark-a-specific-method-or-function-parameter-as-deprecated-in-sphinx

src/rev_ai/apiclient.py
```diff
@@ -184,14 +195,23 @@ def submit_job_local_file(
             Sections of transcript needed to be transcribed.
         :param speaker_names: Only available with "human" transcriber.
             Human readable names of speakers in the file.
+        :param notification_config: CustomerUrlData object containing the callback url to
+            invoke on job completion as a webhook and optional authentication headers to use when
+            calling the callback url
         :returns: raw response data
-        :raises: HTTPError
+        :raises: HTTPError, ValueError
```
amikofalvy, line 197: If you let API validate then you can remove ValueError

HISTORY.rst
```diff
@@ -94,3 +94,8 @@ History
 ------------------
 * Add topic extraction client
 * Add speaker_names to async client for human transcription
+
+2.16.0
+------------------
+* Add source_config and notification_config job options to support customer provided urls with authentication headers
+* Deprecate media_url and callback_url
```
amikofalvy: I would make mention that source config is the replacement for media_url and notification_config is the replacement for callback_url.

README.md
amikofalvy: My previous comment was not accurate. Please disregard.

examples/async_auth_options_example.py
```diff
@@ -0,0 +1,37 @@
...
+# Submitting a job with the authentication header options
+source_url = "https://www.rev.ai/FTC_Sample_1.mp3"
+source_auth_headers = "Authentication: Bearer <token>"
+callback_url = "https://www.example.com/callback"
+callback_auth_headers = "Authentication: Bearer <token>"
+
+# Create config objects containing the url and auth headers for each option
+# These options replace media_url and callback_url and should not be used alongside them
+job = client.submit_job_url(
+    source_config=CustomerUrlData(source_url, source_auth_headers),
```
amikofalvy: I think you should use named parameters (not positional) in your example to add clarity.

src/rev_ai/apiclient.py
amikofalvy: Do you still need the value error?

src/rev_ai/models/customer_url_data.py
```diff
@@ -0,0 +1,19 @@
+class CustomerUrlData:
+    """Object used to contain a url provided by the customer along with any other related information,
+     such as authentication headers."""
+
+    def __init__(self, url, auth_headers=None):
+        """Constructor
+        :param url: customer provided url
+        :param auth_headers: optional customer provided headers to access the url
```
amikofalvy, line 12: I think this need to include type information. Is it a string, an array of strings, or a special auth header object?

amikofalvy (changes requested):
I left a few suggestions for updated comments on python objects.

HISTORY.rst
amikofalvy: I know you mention in the deprecation comment on each method but the changelog should include this information as well.

src/rev_ai/apiclient.py
aaron-wilson-rev: I think so, because we are still validating that they submit a non-null filename in this case. We were raising the ValueError but it wasn't in the doc. We don't need to remove that specific validation for this ticket since it doesn't use source_config or media_urll

src/rev_ai/models/customer_url_data.py
aaron-wilson-rev: Added a description and example

HISTORY.rst
aaron-wilson-rev: Good idea

src/rev_ai/models/customer_url_data.py
```diff
@@ -0,0 +1,22 @@
+class CustomerUrlData:
+    """Object used to contain a url provided by the customer along with any other related information,
+     such as authentication headers."""
+
+    def __init__(self, url, auth_headers=None):
+        """Constructor
+        :param url: customer provided url
+        :param auth_headers: optional customer provided headers to access the url.
+        Only the “Authorization” header is currently supported.
+        This should be a json string in the form '{"<header>": "<scheme> <token>"}'.
```
amikofalvy:
```suggestion
        This should be a stringified json in the form '{"<header>": "<scheme> <token>"}'.
```

aaron-wilson-rev:
So I was wrong about how it expects the auth_headers argument, but I've tested it now with the current input and it works. It should be a dictionary, not a string. It gets converted to json in the request.

aaron-wilson-rev:
Updated to include the LID changes

examples/language_identification_example.py
```diff
@@ -18,18 +18,22 @@
...
 # String containing your access token
+from rev_ai.models.customer_url_data import CustomerUrlData
+
 access_token = "<your_access_token>"
 
 # Create your api client
 client = language_identification_client.LanguageIdentificationClient(access_token)
 
 # Submitting a job with a link to the file you want to identify the language of
 # Change url to your url
```
amikofalvy: I don't think this comment makes sense with the new source_config. We are encouraging people to use source config and notification config instead of  media and callback urls

amikofalvy (approved):

examples/language_identification_example.py
aaron-wilson-rev: Yeah it could be more clear. Updated with `# Change the url specified in source_config to the url of the media to submit`

src/rev_ai/apiclient.py
```diff
@@ -61,18 +61,20 @@ def submit_job_url(
...
         :param metadata: info to associate with the transcription job
-        :param callback_url: callback url to invoke on job completion as
-                             a webhook
-        :param skip_diarization: should Rev AI skip diaization when
-                                 transcribing this file
-        :param skip_punctuation: should Rev AI skip punctuation when
-                                 transcribing this file
+        :param callback_url: callback url to invoke on job completion as a webhook
+        .. deprecated:: 2.16.0
+                Use notification_config instead
```
kbridbur: indentation

src/rev_ai/models/customer_url_data.py
```diff
@@ -0,0 +1,22 @@
+class CustomerUrlData:
```
kbridbur, line 5: needs a 
```
# -*- coding: utf-8 -*-
"""some description"""
```
at the top of the file

kbridbur (approved):

PR #85, opened 2022-04-19
AISDK-192: Add speaker names

kbridbur:


src/rev_ai/models/asynchronous/transcript.py
```diff
@@ -27,35 +27,71 @@ def from_json(cls, json):
 
 
 class Monologue:
-    def __init__(self, speaker, elements):
+    def __init__(self, speaker, speaker_info, elements):
```
amikofalvy: I don't think you should inject new parameters in the middle of the initializer. Theoretically someone could create a Monologue object and this would break their implementation.


can you pass None as default value for the initializer?

tests/test_topic_extraction_client.py
```diff
@@ -14,7 +14,7 @@
 
 TOKEN = 'token'
 TEXT = 'Input'
-JSON = Transcript([Monologue(0, [Element('text', 'hello', 0.0, 0.1, 100)])])
+JSON = Transcript([Monologue(0, None, [Element('text', 'hello', 0.0, 0.1, 100)])])
```
amikofalvy: `None` should come at the end now.

amikofalvy (approved):

PR #91, opened 2022-05-03
AISDK-197: Add sentiment analysis client

kbridbur:


examples/sentiment_analysis_example.py
```diff
@@ -0,0 +1,90 @@
+"""Copyright 2019 REV
```
amikofalvy: Year should be 2022
```suggestion
"""Copyright 2022 REV
```

examples/sentiment_analysis_example.py
```diff
@@ -0,0 +1,90 @@
...
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
amikofalvy, line 3: Why is this the Apache license vs the MIT license?

examples/sentiment_analysis_example.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+# String containing your access token
+access_token = "<your_access_token>"
+
+# Create your api client
+client = sentiment_analysis_client.SentimentAnalysisClient(access_token)
+
+# Submit a job with whatever text you want by changing this input
+text = "<input_text>"
```
amikofalvy: I think we should seed with some example text. The example should require as little input during the example run as possible.

src/rev_ai/__init__.py
```diff
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 """Top-level package for rev_ai"""
 
-__version__ = '2.15.1'
+__version__ = '2.15.2'
```
amikofalvy: make sure to coordinate your version with Aaron's PR

src/rev_ai/models/insights/insights_job.py
```diff
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 """Job model"""
 
-from ...asynchronous.job_status import JobStatus
+from ..asynchronous.job_status import JobStatus
 
 
-class TopicExtractionJob:
+class InsightsJob:
```
amikofalvy: I'm not sure how I feel about this change. They really are two different APIs that will evolve independently and the products are sold separately (further reinforcing the idea that they are separate).

It wouldn't be unreasonable to see how they could start to diverge in the future. Calling both Insights Jobs is a step in a different direction.

kbridbur: Thats a fair point. I'll keep them separate

examples/sentiment_analysis_example.py
kbridbur: Seems reasonable. I'll use some text from wikipedia or something

examples/sentiment_analysis_example.py
kbridbur: Our examples are all under apache. I don't know the distiction. I don't think it really matters

src/rev_ai/models/insights/sentiment_analysis/sentiment_value.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+"""Enum for possible sentiments"""
+
+from enum import Enum
+
+
+class SentimentValue(Enum):
```
beaudrychase, line 7: should this be a union type?

kbridbur: I don't believe union types don't exist in python

src/rev_ai/models/insights/sentiment_analysis/sentiment_value.py
```diff
@@ -0,0 +1,17 @@
...
+    POSITIVE = 1
+    NEGATIVE = 2
+    NEUTRAL = 3
```
beaudrychase, line 10:
```suggestion
type Sentiment = 'positive' | 'negative' | 'neutral';
```

beaudrychase (approved):
looks good to me. Pretty similar to topic

src/rev_ai/sentiment_analysis_client.py
```diff
@@ -0,0 +1,97 @@
...
+
+        GenericApiClient.__init__(self, access_token, self.api_name, self.api_version,
+                                  SentimentAnalysisJob.from_json, SentimentAnalysisResult.from_json)
+
+    def submit_job_from_text(self,
+                             text=None,
+                             metadata=None,
+                             callback_url=None,
+                             delete_after_seconds=None,
+                             language=None):
```
aaron-wilson-rev: Now that I merged AISDK-190 this should have notification_config as in the other features

examples/sentiment_analysis_example.py
```diff
@@ -0,0 +1,97 @@
...
+    rain, with parasol used when protecting oneself from sunlight, though the terms continue to be \
+    used interchangeably. Often the difference is the material used for the canopy; some parasols \
+    are not waterproof, and some umbrellas are transparent. Umbrella canopies may be made of \
+    fabric or flexible plastic. There are also combinations of parasol and umbrella that are \
+    called en-tout-cas (French for 'in any case')."
+job = client.submit_job_from_text(text,
+                                  metadata=None,
+                                  callback_url=None,
+                                  delete_after_seconds=None,
+                                  language=None)
```
aaron-wilson-rev: also should show notification_config=none here when that's added

aaron-wilson-rev (changes requested):
Should support the notification_config option now that the others do

tests/test_generic_api_client.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 import sys
 
-from rev_ai.models.customer_url_data import CustomerUrlData
+from src.rev_ai.models.customer_url_data import CustomerUrlData
```
kbridbur, line 7: @aaron-wilson-rev a lot of these paths were giving me import errors in my ide. Seems like tests were passing on your branch though. Not sure why

aaron-wilson-rev: I have the source root in my python path, I'm guessing the CI job does too and maybe you don't?

src/rev_ai/models/__init__.py
```diff
@@ -5,5 +5,7 @@
 from .streaming import MediaConfig
 from .asynchronous import Job, JobStatus, Account, Transcript, Monologue, Element, CaptionType, \
     SpeakerName
-from .insights import TopicExtractionJob, TopicExtractionResult, Topic, Informant
+from .insights import TopicExtractionJob, TopicExtractionResult, Topic, Informant, \
+    SentimentAnalysisResult, SentimentValue, SentimentMessage, SentimentAnalysisJob
 from .language_id import LanguageIdentificationJob, LanguageIdentificationResult, LanguageConfidence
+from .customer_url_data import CustomerUrlData
```
kbridbur, line 11: @aaron-wilson-rev was it intentional to not add it here? We generally propagate all models up to the top level package so customers only need 1 import statement

tests/test_generic_api_client.py
aaron-wilson-rev: Looks like it does not break anything to have the full path regardless, so probably better to have that anyway

src/rev_ai/models/__init__.py
aaron-wilson-rev: No just an oversight

aaron-wilson-rev (approved):

PR #93, opened 2022-05-17
AISDK-206: Add Language Parameter to streaming client

kshiflett88:


src/rev_ai/models/streaming/mediaconfig.py
```diff
@@ -28,4 +30,5 @@ def get_content_type_string(self):
             (';layout={}'.format(self.layout) if self.layout else '') + \
             (';rate={}'.format(self.rate) if self.rate else '') + \
             (';format={}'.format(self.format) if self.format else '') + \
-            (';channels={}'.format(self.channels) if self.channels else '')
+            (';channels={}'.format(self.channels) if self.channels else '') + \
+            (';language={}'.format(self.language) if self.language else '')
```
amikofalvy: This does not work. The language parameter is not part of the content_type parameter.

`language` needs to be it's own query parameter

amikofalvy (changes requested):
The approach does not work

amikofalvy (changes requested):
Approach looks good. We need to bump the version number for release.

src/rev_ai/models/streaming/mediaconfig.py
kshiflett88: > Approach looks good. We need to bump the version number for release.

@amikofalvy for clerification do I change the `current_version` in `setup.cfg` from 2.16.0 to 2.17.0 and then add a note in the `HISTORY.rst` file about the change to the version?

amikofalvy (approved):

amikofalvy:
After you merge this you will need to add a changelog entry on the docs site as well. Like this one for the NodeSDK https://github.com/revdotcom/revai-docs/pull/141

PR #95, opened 2022-06-02
AISDK-221 Fix HIPAA Error Message on Incorrect Usage

aaron-wilson-rev:
The python sdk does not try to wrap any of the http error responses from the API, so no need to fix handling of a 403 here. This PR just removes some unnecessary test entries.

### DIFF
--- a/tests/helpers/errors.py
+++ b/tests/helpers/errors.py
@@ -7,6 +7,10 @@ ERRORS = {
         "title": "Authorization has been denied for this request",
         "status": 401
     },
+    'forbidden': {
+        "status": 403,
+        "error": "HIPAA user is not allowed to use MediaUrl"
+    },
     'job-not-found': {
         "type": "https://www.rev.ai/api/v1/errors/job-not-found",
         "title": "could not find job",
--- a/tests/helpers/errors.py
+++ b/tests/helpers/errors.py
@@ -14,7 +14,7 @@ ERRORS = {
     },
     'forbidden': {
         "status": 403,
-        "error": "HIPAA user is not allowed to use MediaUrl"
+        "error": "The client is forbidden to submit that request"
     },
     'job-not-found': {
         "type": "https://www.rev.ai/api/v1/errors/job-not-found",
--- a/tests/helpers/errors.py
+++ b/tests/helpers/errors.py
@@ -3,6 +3,11 @@ def get_error_test_cases(errors):
 
 
 ERRORS = {
+    'invalid-parameters': {
+        "type": "https://www.rev.ai/api/v1/errors/invalid-parameters",
+        "title": "Your request parameters did not validate",
+        "status": 400
+    },
     'unauthorized': {
         "title": "Authorization has been denied for this request",
         "status": 401
--- a/tests/helpers/errors.py
+++ b/tests/helpers/errors.py
@@ -3,16 +3,6 @@ def get_error_test_cases(errors):
 
 
 ERRORS = {
-    'invalid-parameters': {
-        "parameter": {
-            "example": [
-                "The example field is required"
-            ]
-        },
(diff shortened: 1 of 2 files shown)


tests/helpers/errors.py
```diff
@@ -3,16 +3,6 @@ def get_error_test_cases(errors):
 
 
 ERRORS = {
-    'invalid-parameters': {
```
dmtrrk, line 6: why did you remove this?

aaron-wilson-rev: It's not used

tests/helpers/errors.py
```diff
@@ -22,13 +12,6 @@ def get_error_test_cases(errors):
         "title": "could not find job",
         "status": 404
     },
-    'out-of-credit': {
-        "title": "You do not have enough credits",
-        "type": "https://www.rev.ai/api/v1/errors/out-of-credit",
-        "detail": "You have only 60 seconds remaining",
-        "current_balance": 60,
-        "status": 403
-    },
```
eugenep-rev, line 31: can you explain the fix? why just removing fixes the issue?

aaron-wilson-rev: there isn't actually a problem with the python sdk since it just passes through the api error. I just removed these from our test resources since they are never used in our tests.

tests/helpers/errors.py
dmtrrk: I think, we might need to cleanup [tests](https://github.com/revdotcom/revai-python-sdk/blob/develop/tests/test_baseclient.py#L31) if we are deleting test cases

tests/helpers/errors.py
```diff
@@ -4,31 +4,23 @@ def get_error_test_cases(errors):
 
 ERRORS = {
     'invalid-parameters': {
-        "parameter": {
-            "example": [
-                "The example field is required"
-            ]
-        },
         "type": "https://www.rev.ai/api/v1/errors/invalid-parameters",
-        "title": "Your request parameters didn't validate",
```
aaron-wilson-rev, line 13: this breaks regex on the test verification so easiest to just cut it out. I'm guessing this is why it was left out

tests/helpers/errors.py
aaron-wilson-rev: this error wasn't actually being tested before, so instead of removing it I fixed it and included it in the tests to make sure we raise a 400 response properly

tests/helpers/errors.py
```diff
@@ -4,31 +4,23 @@ def get_error_test_cases(errors):
...
         "status": 400
     },
     'unauthorized': {
         "title": "Authorization has been denied for this request",
         "status": 401
     },
+    'forbidden': {
+        "status": 403,
+        "error": "HIPAA user is not allowed to use MediaUrl"
+    },
```
eugenep-rev, line 18: I"m confused about this PR. Looks like we change tests only here. Also, we return 403 when we use wrong language or human transcriber.

aaron-wilson-rev: Yes, because we have test cases that today are not used in the test and are misleading. I'll make the 403 test more generic. The main point of this test is just that we raise the error if we get one back from the api.

tests/helpers/errors.py
dmtrrk: correct me if I'm wrong, but in the test you're trying to get a test data by the provided key in `@pytest.mark.parametrize`. and once you deleted test data, this function might return `None` and it might be the test is not failing eventually

aaron-wilson-rev: 'invalid-parameters' was not included in the parameters before

aaron-wilson-rev: But now I added it to the parameters and restored it in the errors file so we can test it too.

dmtrrk (approved):

PR #96, opened 2022-06-10
AISDK-189: Update all sdk's to have the option to submit to eu endpoint 

jennywong2129:


src/rev_ai/models/asynchronous/transcript_type.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+"""Enum for transcript content types"""
+
+from enum import Enum
```
jennywong2129, line 4: @amikofalvy was there a reason we dint have this enum class but have one for captions

PR #97, opened 2022-07-24
AISDK-229: Add skip_postprocessing option

kbridbur:


src/rev_ai/streamingclient.py
```diff
@@ -72,7 +72,8 @@ def start(self,
               detailed_partials=None,
               start_ts=None,
               transcriber=None,
-              language=None):
+              language=None,
+              skip_postprocessing=None):
```
dmtrrk, line 76: Why is it None rather than False?

dmtrrk (approved):
It doesn't look like we need None as a value, but False for the new parameter. If no changes are required, this looks good

src/rev_ai/streamingclient.py
kbridbur: I did it mostly for consistency with the other boolean parameters. The reasoning I see for making it None is that it's not the SDKs place to assume what the API defaults are so it should default to all parameters being empty

src/rev_ai/apiclient.py
```diff
@@ -109,6 +110,7 @@ def submit_job_url(
         :param notification_config: CustomerUrlData object containing the callback url to
             invoke on job completion as a webhook and optional authentication headers to use when
             calling the callback url
+        :param skip_postprocessing: skip all text postprocessing (punctuation, capitalization, ITN)
```
beaudrychase, line 113: What do you think about writing out ITN? Is this a commonly known acronym?

tests/test_job.py
```diff
@@ -129,7 +130,7 @@ def test_submit_job_url_with_success(self, mock_session, make_mock_response):
                                     NOTIFICATION_URL, True,
                                     True, 1, CUSTOM_VOCAB, True,
                                     True, 0, LANGUAGE, CUSTOM_VOCAB_ID,
-                                    TRANSCRIBER)
+                                    TRANSCRIBER, skip_postprocessing=True)
```
beaudrychase, line 133:
```suggestion
                                    TRANSCRIBER, True)
```
This better matches the pattern of the other parameters

tests/test_job.py
```diff
@@ -288,7 +294,7 @@ def test_submit_job_local_file_with_success(self, mocker, mock_session, make_moc
                                                NOTIFICATION_URL, True,
                                                True, 1, CUSTOM_VOCAB, True,
                                                True, 0, LANGUAGE, CUSTOM_VOCAB_ID,
-                                               TRANSCRIBER)
+                                               TRANSCRIBER, skip_postprocessing=True)
```
beaudrychase, line 297:
```suggestion
                                               TRANSCRIBER, True)
```
same here

beaudrychase (approved):
left two minor style suggestions. Looks good to me otherwise

src/rev_ai/apiclient.py
hbkse: I agree that ITN is confusing

tests/test_job.py
kbridbur: There are other parameters before skip_postprocessing which aren't provided. In order to avoid writing everything out this was needed

PR #104, opened 2023-11-02
Add support for diarization_type, speakers_count, remove_atmospherics

hrist0stoichev:
This PR fixes the support for `speakers_count` and `remove_atmospherics`, and adds support for `diarization_type`.

PR #105, opened 2023-12-20
REVAI-3855: Update python sdk to support Super API

kirillatrev:
REVAI-3855: Update python sdk to support Super API
https://revinc.atlassian.net/browse/REVAI-3855


kirillatrev:
Tests are coming shortly

src/rev_ai/apiclient.py
```diff
@@ -451,6 +467,160 @@ def get_account(self):
...
+        :param id_: id of job to be requested
+        :returns: transcript data as text
+        :raises: HTTPError
+        """
+        if not id_:
+            raise ValueError('id_ must be provided')
+
+        response = self._make_http_request(
+            "GET",
+            urljoin(self.base_url, 'jobs/{}/transcript/summary'.format(id_)),
```
k-weng, line 534: Any way to extract `'jobs/{}/transcript/summary'` to be reusable?

src/rev_ai/models/asynchronous/job.py
```diff
@@ -28,7 +29,9 @@ def __init__(
             rush=None,
             segments_to_transcribe=None,
             remove_atmospherics=None,
-            speakers_count=None):
+            speakers_count=None,
+            summarization: Summarization = None,
+            translation: Translation = None):
```
k-weng, line 34:
```suggestion
            speakers_count=None,
            summarization: Summarization=None,
            translation: Translation=None):
```

src/rev_ai/models/asynchronous/summarization_options.py
```diff
@@ -0,0 +1,63 @@
...
+"""Summarization request options."""
+
+
+class SummarizationOptions:
+    def __init__(
+            self,
+            prompt: str = None,
+            model: NlpModel = None,
+            formattingType: SummarizationFormattingOptions = None
+    ):
```
k-weng: Is this formatting correct?

tests/resources/test_mp3.mp3
k-weng, line 1: Do we need these files? Can we just mock the files like we do in test_job?

src/rev_ai/models/asynchronous/summarization_options.py
kirillatrev: I used auto formatter as suggested with linter, no complaints.

tests/test_superapi.py
k-weng, line 1: I would split the unit tests up into simpler test cases. They are pretty hefty, each testing three or four response methods along with the request method. I would maybe split each test case into
1. request_local_file - assert request with proper params
2. request_source_url - assert request with proper params
3. get_transcript_summary_text
4. get_transcript_summary_json
5. get_transcript_summary_streamed_json
6. get_translation...
...

k-weng (changes requested):
Looking good so far - had some comments regarding simplifying the unit tests and the need for the audio files and transcript file since they can just be mocked in code. 

In addition, it would nice to add an example to https://github.com/revdotcom/revai-python-sdk/tree/develop/examples and document some usage in the README - like we have for Human Transcription - https://github.com/revdotcom/revai-python-sdk?tab=readme-ov-file#human-transcription

src/rev_ai/models/asynchronous/summarization_options.py
kirillatrev: I follow PEP8, which is commonly accepted standard for python formatting, PyCharm supports PEP 8 and suggests this formatting here. I Used autoformat everywhere where reasonable.

k-weng:
Can you also bump version to `2.19.0` and add an entry to https://github.com/revdotcom/revai-python-sdk/blob/develop/HISTORY.rst

src/rev_ai/models/asynchronous/job.py
kirillatrev: Same as bellow:

```
I follow PEP8, which is commonly accepted standard for python formatting, PyCharm supports PEP 8 and suggests this formatting here. I Used autoformat everywhere where reasonable.

File: tests/resources/test_mp3.mp3
Comment (kirillatrev) [reply tests/resources/test_mp3.mp3:1]: I believe they are mocked now, so I will remove them from PR. Thanks!

File: tests/test_superapi.py
Comment (kirillatrev) [reply tests/test_superapi.py:1]: This used to be a real integration test before, but as of now I agree that it makes more sense to split those tests. Thanks!

File: src/rev_ai/apiclient.py [hunk 4]
```diff
@@ -66,7 +70,9 @@ def submit_job_url(
             notification_config=None,
             skip_postprocessing=False,
             remove_atmospherics=False,
-            speakers_count=None):
+            speakers_count=None,
```
dmtrrk, line 73: it looks like we are missing to pass these parameters to `_create_job_options_payload`
```
remove_atmospherics=False,
speakers_count=None,
```
@amikofalvy do you know whether it is intentional?

src/rev_ai/apiclient.py
```diff
@@ -397,6 +413,27 @@ def get_captions(self, id_, content_type=CaptionType.SRT, channel_id=None):
...
+        :returns: caption data as text
+        :raises: HTTPError
+        """
+        if not id_:
+            raise ValueError('id_ must be provided')
+        query = self._create_captions_query(channel_id)
+
+        response = self._make_http_request(
+            "GET",
+            urljoin(self.base_url, 'jobs/{0}/captions/translation/{2}{1}'.format(id_, query, language)),
```
dmtrrk: why don't you pass params as `format(id_, language, query)`?
it is a bit difficult to read the code with the unnecessary `{2}{1}`

src/rev_ai/models/asynchronous/job.py
dmtrrk: I think it is less matter what is the autoformatting when all the remaining code will be formatted differently.

at the same time, as per [PEP8](https://peps.python.org/pep-0008/#other-recommendations)

```
Don’t use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter:
# Correct:
def complex(real, imag=0.0):
    return magic(r=real, i=imag)
# Wrong:
def complex(real, imag = 0.0):
    return magic(r = real, i = imag)
```

src/rev_ai/models/asynchronous/summarization_options.py
```diff
@@ -0,0 +1,63 @@
+from .summarization_formatting_options import SummarizationFormattingOptions
+from .summarization_job_status import SummarizationJobStatus
+from ..nlp_model import NlpModel
+
+"""Summarization request options."""
```
dmtrrk: I think this should be inside the class

src/rev_ai/models/asynchronous/summarization_options.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class SummarizationOptions:
+    def __init__(
```
dmtrrk, line 9: based on other files, it looks like the trailing `):` should be on the same line with the last argument.

nit, but this is a public code. also, there should be a comment

src/rev_ai/models/asynchronous/summarization_options.py
```diff
@@ -0,0 +1,63 @@
...
+            dict_result['prompt'] = self.prompt
+        if self.model:
+            dict_result['model'] = self.model
+        if self.type:
+            dict_result['type'] = self.type
+
+        return dict_result
+
+
+"""Summarization options."""
```
dmtrrk: not sure what kind of syntax is this. I believe this should be inside the class

src/rev_ai/models/asynchronous/summary.py
```diff
@@ -0,0 +1,20 @@
+from typing import List, Dict
+
+"""Transcript summary model."""
```
dmtrrk: this [should be](https://peps.python.org/pep-0008/#:~:text=Imports%20are%20always%20put%20at,before%20module%20globals%20and%20constants.) either before imports, or inside the class

```
Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
```


so the file might look like this:
```
"""Module description here"""

import xyz

class MyClass:
    """This is my class"""
    pass

class MyClass2:
    """This is my class 2"""
    pass
```

src/rev_ai/models/asynchronous/translation_language_options.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class TranslationLanguage(TranslationLanguageOptions):
+    def __init__(
+            self,
+            language: str = None,
+            model: NlpModel = None,
+            status: SummarizationJobStatus = None,
+            failure: str = None
+
```
dmtrrk: remove line

src/rev_ai/models/asynchronous/translation_language_options.py
```diff
@@ -0,0 +1,53 @@
...
+from ..nlp_model import NlpModel
+
+"""Translation language request options."""
+
+
+class TranslationLanguageOptions:
+    def __init__(
+            self,
+            language: str = None,
+            model: NlpModel = None
```
dmtrrk: not sure whether you discussed it already., I see that some code is written in py2.x style without the type annotations, other - with..

while strict typing is good, wondering how can we be more consistent in the publicly available code

dmtrrk (changes requested):

src/rev_ai/apiclient.py
kirillatrev: The idea was to avoid adding changes to format arguments order, but at this point it is better to change it, Thanks!

tests/test_async_summarization.py
```diff
@@ -0,0 +1,214 @@
...
+                                        "Try to summarize this transcript as good as you possibly can",
+                                        NlpModel.PREMIUM,
+                                        SummarizationFormattingOptions.BULLETS
+
+                                    ))
+        mock_session.request.assert_called_once_with(
+            "POST",
+            JOBS_URL,
+            json={
+                'metadata': "python sdk SuperApi test",
```
dmtrrk: not sure `SuperApi` should be here

src/rev_ai/models/asynchronous/job.py
kirillatrev: Discussed on Zoom. Agreed that formatting of arguments with type annotation is different according to the standard.

kirillatrev:
Examples to follow in another PR once this update is published

src/rev_ai/models/asynchronous/translation_language_options.py
kirillatrev: Agreed to encourage incremental improvements going forward.

src/rev_ai/apiclient.py
k-weng: @dmtrrk We just haven't supported them in SDKs yet - there's an open PR for it -https://github.com/revdotcom/revai-python-sdk/pull/104

k-weng (approved):
LGTM - I'll let @dmtrrk provide the final approval

dmtrrk (approved):

PR #106, opened 2023-12-23
REVAI-3855: examples

kirillatrev:


examples/async_summarize.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Checks if the job has been transcribed and summarized. Please note that this is not the recommended way
+    # of getting job status in a real application. For recommended methods of getting job status
+    # please see our documentation on setting a callback url here:
+    # https://docs.rev.ai/resources/tutorials/get-started-api-webhooks/
+    if status == JobStatus.IN_PROGRESS or summarization_status == SummarizationJobStatus.IN_PROGRESS:
+        time.sleep(5)
+        continue
+
+    elif status == JobStatus.FAILED or summarization_status == SummarizationJobStatus.FAILED:
```
dmtrrk:
```suggestion
    if status == JobStatus.FAILED or summarization_status == SummarizationJobStatus.FAILED:
```

examples/async_summarize.py
```diff
@@ -0,0 +1,94 @@
...
+        print("Job Failed : {}".format(job_details.failure_detail))
+        break
+
+    if status == JobStatus.TRANSCRIBED and summarization_status == SummarizationJobStatus.COMPLETED:
+        # Getting a list of current jobs connected with your account
+        # The optional parameters limits the length of the list.
+        # starting_after is a job id which causes the removal of
+        # all jobs from the list which were created before that job
+        list_of_jobs = client.get_list_of_jobs(limit=None, starting_after=None)
```
dmtrrk: Do we need this call in this example

tests/test_async_summarization.py
```diff
@@ -141,7 +141,24 @@ def test_get_transcript_summary_text(self, mock_session, make_mock_response):
             headers=self.hdr_fixture(client, {'Accept': 'text/plain'})
         )
 
-    def test_get_transcript_summary_json_paragraph(self, mock_session, make_mock_response):
+    def test_get_transcript_summary_object(self, mock_session, make_mock_response):
```
dmtrrk, line 144: It looks like we are missing a test for json method now, no?

dmtrrk (approved):
A few minor suggestions

examples/async_summarize.py
kirillatrev: Just curious, why do you think this way is better? All examples follow the same pattern...

tests/test_async_summarization.py
kirillatrev: Both are there, it's just a weird diff view

examples/async_summarize.py
dmtrrk: The same reason you don't use `else` for the next block. It is redundant. Having both elif is ok as well. Having them different requires to check why is it different.. 
The current code is correct as well

dmtrrk (approved):

examples/async_summarize.py
```diff
@@ -0,0 +1,97 @@
...
+
+# Submitting a job with a link to the file you want transcribed
+# Change url to your url, custom_vocabularies is optional like above
+url = "https://www.rev.ai/FTC_Sample_1.mp3"
+job = client.submit_job_url(media_url=url,
+                            delete_after_seconds=2592000,
+                            language='en',
+                            summarization_config=SummarizationOptions(
+                                model=NlpModel.PREMIUM
+                            ))
```
alexsku: this is odd formatting style (the integration instead of 4 spaces is much larger) are you trying to follow the existing format?

kirillatrev: Most likely linter was complaining about the line length. I'll double check.

kirillatrev: In general - yes I was trying to follow existing formatting in other example files for consistency

kirillatrev:
@alexsku good to merge?

alexsku:
sure?

dmtrrk (approved):

alexsku (approved):
:shipit:

PR #113, opened 2024-06-17
Add support for diarization_type, speakers_count, remove_atmospherics

kirillatrev:
Add support for diarization_type, speakers_count, remove_atmospherics

Follow up on 
https://github.com/revdotcom/revai-python-sdk/pull/104


src/rev_ai/apiclient.py
```diff
@@ -137,6 +139,10 @@ def submit_job_url(
                                                    segments_to_transcribe, speaker_names,
                                                    source_config, notification_config,
                                                    skip_postprocessing,
+						   remove_atmospherics,
```
dmtrrk, line 142: is this correct you added params except `diarization_type`?

kirillatrev: It appears that they were declared on the prototype but were not actually submitted....

dmtrrk: @eugenep-rev do you know anything about this?

I just can guess that was intentionally, but looks like a bug

kirillatrev:
@hrist0stoichev hrist0stoichev thank you for your contribution.

eugenep-rev (approved):

PR #116, opened 2024-10-24
DOCS-298: Update Rev AI python sdk to support different base url

jennywong2129:


src/rev_ai/apiclient.py
```diff
@@ -30,20 +31,25 @@ class RevAiAPIClient(BaseClient):
...
         :param access_token: access token which authorizes all requests and links them to your
                              account. Generated on the settings page of your account dashboard
                              on Rev AI.
+        :param url (optional): url of the Rev AI API deployment to use, defaults to the US
+                    deployement, i.e. 'https://api.rev.ai', which can be referenced as
+                    RevAiApiDeploymentConfigMap[RevAiApiDeployment.US]['base_url'].
         """
 
+        # Default speech to text base url
+        self.base_url = '{0}/speechtotext/{1}/'.format(url if url else self.default_url, self.version)
```
alexsku: minor but why not using f strings?

src/rev_ai/generic_api_client.py
```diff
@@ -13,7 +14,10 @@ class GenericApiClient(BaseClient):
     """Generic client which handles logic for making requests to almost any Rev AI Api.
     Intended to be inherited and extended by a specific client per API"""
 
-    def __init__(self, access_token, api_name, api_version, parse_job_info, parse_job_result):
+    # Default url for US Rev AI deployment
+    default_url = RevAiApiDeploymentConfigMap[RevAiApiDeployment.US]['base_url']
+    
+    def __init__(self, access_token, api_name, api_version, parse_job_info, parse_job_result, url=None):
```
alexsku: i would recommend using type hints whenever is possible

src/rev_ai/language_identification_client.py
```diff
@@ -20,17 +20,20 @@ class LanguageIdentificationClient(GenericApiClient):
...
                              on Rev AI.
+        :param url (optional): url of the Rev AI API deployment to use, defaults to the US
+                    deployement, i.e. 'https://api.rev.ai', which can be referenced as
+                    RevAiApiDeploymentConfigMap[RevAiApiDeployment.US]['base_url'].
         """
 
         GenericApiClient.__init__(self, access_token, self.api_name, self.api_version,
                                   LanguageIdentificationJob.from_json,
-                                  LanguageIdentificationResult.from_json)
+                                  LanguageIdentificationResult.from_json, url)
```
alexsku, line 37: i'm not sure i understand the pattern used for whether a parameter on a new line or not, are you just fitting as many in a line as possible? this is fine, i personally i think have each parameter on a new line but again i don't think this is important either way

src/rev_ai/models/revaiapi_deployment_config_constants.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+class RevAiApiDeployment(Enum):
```
alexsku, line 5: not super critical but perhaps we should be adding docstrings for such things

src/rev_ai/models/revaiapi_deployment_config_constants.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+class RevAiApiDeployment(Enum):
+    US = 'US'
+    EU = 'EU'
+
+RevAiApiDeploymentConfigMap = {
+    RevAiApiDeployment.US: { 'base_url': 'https://api.rev.ai', 'base_websocket_url': 'wss://api.rev.ai' },
```
alexsku: should the value be a TypedDict? and i would use type hints here as well

alexsku (approved):
left some stylistics comments, then :shipit:

src/rev_ai/models/revaiapi_deployment_config_constants.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+class RevAiApiDeployment(Enum):
+    US = 'US'
+    EU = 'EU'
+
+RevAiApiDeploymentConfigMap = {
```
eugenep-rev: just curious, why do we need this map? why just not follow usual pattern with options classes?

```
RevAiStreamingClient(..., RevAiStreamingClientOptions(....))
```

we can even define a few static preconfigured deployments: RevAiStreamingClientOptions.US, RevAiStreamingClientOptions.EU and pass them or something like this.

eugenep-rev (approved):

src/rev_ai/apiclient.py
jennywong2129: sure

src/rev_ai/language_identification_client.py
jennywong2129: i think the pattern is fitting the most in a line

src/rev_ai/models/revaiapi_deployment_config_constants.py
jennywong2129: i want to reuse this for any type of revai service, not just streaming

